PR #1, opened 2021-05-20
Setup API Extractor and API Documenter

Devessier:
We installed @microsoft/api-extractor and @microsoft/api-documenter to automatically generate a documentation from all of our packages.
This documentation is published on Github Pages.

And this is so cool 😎.


prastoin (approved):
Great, docs everywhere perfect theme choice

PR #2, opened 2021-05-21
Set up react native client

prastoin:


Devessier (approved):
Nice setup!
We can now build the clients on solid foundations.

PR #10, opened 2021-05-28
Create basic chat

### LINKED ISSUE #9: Set up a type-safe socket connection between the front-end and the back-end
We will need socket connections for several different purposes, all around our application:

- Chats (the most obvious purpose)
- Live and Collaborative Playlist Editors
- Track Vote Room

We should guarantee type-safe exchanges to ease development and to have a more predictable code.

Devessier:
We established a type-safe socket connection between the front-end and the back-end, using Socket.IO.
We must write Socket.IO event handlers in `start/socket.ts` file and controllers in `app/Controllers/Ws` directory. We call these controllers in socket handlers.
Socket controllers must expose public static methods, that conform to `WsControllerMethodArgs` interface.

We still have to handle authentication through sockets, which can be a bit bothering.

Closes #9


prastoin (dismissed):
Really nice !
Some super cool state machine on the horizon
Thanks for the search refacto on server

prastoin:
Having a persistent build issue with:
`error An unexpected error occurred: "ENOENT: no such file or directory, lstat '/home/runner/work/MusicRoom/MusicRoom/packages/types/node_modules/semver'".`
Seeing later how and why

prastoin:
Edit: On this other way this PR successfully achieved. Then we could conclude the github actions deps issue comes from one of the new packages. Such as socket-io ?

prastoin (dismissed):

prastoin:
Found a workaround in [this issue](https://github.com/yarnpkg/yarn/issues/2629#issuecomment-283580308)
Until we find a best solution ?

Devessier:
We encountered a very strange bug since commit [87d8](https://github.com/AdonisEnProvence/MusicRoom/pull/10/commits/87d810071736d3a08b45c60928b9980b4ad1a8c0). We could not install dependencies as yarn told us that `lstat` failed to find `semver` package in the `node_modules` directory of `types` package.
After several hours of debugging we found that the issue was due to yarn workspaces `nohoist` used in client `package`. When we were removing the `nohoist`, yarn could install the dependencies. But `expo` could not start.
Therefore we had to find how to make expo work in a monorepo without using `nohoist` mechanism.

We had to follow the indications there: [https://github.com/expo/expo/tree/master/packages/expo-yarn-workspaces](https://github.com/expo/expo/tree/master/packages/expo-yarn-workspaces).
We had to be sure to [use the same version of metro-config as react native](https://forums.expo.io/t/javascript-bundle-error-with-transformer-transform-is-not-a-function-after-upgrading-to-expo-sdk-38-and-expo-cli-3-22-3/40822). We had to find in our yarn.lock the version of metro-config and use it.

prastoin (approved):

PR #18, opened 2021-05-31
Set up server testing

### LINKED ISSUE #11: Set up server testing
AdonisJS creator wrote [an article about Testing and Adonis](https://docs.adonisjs.com/cookbooks/testing-adonisjs-apps).
He uses a test runner for Node.js he wrote himself, [Japa](https://github.com/thetutlage/japa).

Japa is more focus on back-end testing issues than Jest, that was primary used to test front-end code.
It intentionally does not support parallel test suites, which can cause unpredictable side-effects when dealing with shared resources, such as a database.
It does not support snapshots, which should be discarded in favour of more user-centric testing strategies (integration testing).

We should give a try to Japa as it's the official way to go, and if we need more than it can offer, we can still replace it with Jest.

Devessier:
We configured japa, installed supertest to send request to our routes, we set up nodemon to listen to changes in all TypeScript files and we configured nyc to collect test coverage.

We still have to enable database testing. We will do it when we will have configured a database.

Closes #11

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,3 +48,4 @@ jobs:
               env:
                 GOOGLE_API_KEY: ${{ secrets.GOOGLE_API_KEY }}
                 HOST: "0.0.0.0"
+                APP_KEY: "MusicRoom"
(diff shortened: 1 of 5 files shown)


prastoin (dismissed):
Great :unicorn:

prastoin (approved):
great :unicorn:

PR #24, opened 2021-06-01
Setup redis server

prastoin:
What have been done here:
-Set up redis container using docker-compose
-Used [adonis first party package](https://docs.adonisjs.com/guides/redis) for redis to establish the connection to the container
-Installed and set up socket-io-redis to be able to emit/broadcast... in multi nodes using socket io


Devessier (approved):
We will be able to scale horizontally our server now that Redis handles Socket.IO connections on different hosts.

PR #25, opened 2021-06-04
Setup temporal io pkg

prastoin:
Needs to be done:

- [x] Use Local Activity to Notify connection between adonis and temporal
- [x] Ci implem
- [x] Temporal tests
- [ ] Room eviction


packages/server/app/Services/Ws.ts
```diff
@@ -1,15 +1,12 @@
...
-} from '@musicroom/types';
-import { createAdapter } from '@socket.io/redis-adapter';
+import { AllEvents } from '@musicroom/types/src/websockets';
+import { createAdapter, RedisAdapter } from '@socket.io/redis-adapter';
 import { RedisClient } from 'redis';
 import { Server } from 'socket.io';
 
 class Ws {
-    public io: Server<ChatClientToServerEvents, ChatServerToClientEvents>;
+    public io: Server<AllEvents>;
```
Devessier: We should separate messages sent by the client to the server and the messages sent by the server to the client.
That is, use two separate types and provide them as two different generics.

Devessier (approved):
Great!
Now Temporal is set up, we can broadcast to users that they have joined a room and the fron-end, thanks to state machines, handles the logic perfectly.

Temporal is such pleasant 👀.

PR #27, opened 2021-06-07
Create simple music track vote

Devessier:


prastoin:
We still need to test if it's still correctly working on IOS as now it's ope on Android

prastoin (approved):

PR #34, opened 2021-06-11
Set up front end testing

Devessier:
We set up Expo Jest, Testing Library React Native and MSW.
It was a bit difficult to set up MSW, as it is meant to be used with Jest inside a `JSDOM` environment. We had to mock the `Fetch` API and the `LocalStorage` API.

We wrote our first integration test, and we mocked `/search/track/:query` API route using MSW.

We listed all the directories containing TypeScript code inside the `tsconfig.json` file.

We set up volta on all the packages. Before the correct version was used only when we were in the top directory.


prastoin (approved):
:dart: Testing like a boss

PR #35, opened 2021-06-11
Connect play pause

prastoin:


packages/client/contexts/MusicPlayerContext.tsx
```diff
@@ -1,42 +1,40 @@
...
-}
-
 type MusicPlayerContextProviderProps = {
     socket: Socket;
 };
 
+//FIXME perfs optimizations here
 export const MusicPlayerContextProvider: React.FC<MusicPlayerContextProviderProps> =
     ({ socket, children }) => {
+        console.log('1');
```
Devessier, line 26: We should remove the console.log calls before merging the PR.

packages/temporal/activities.go
```diff
@@ -24,16 +24,19 @@ func PingActivity(_ context.Context) error {
 }
 
 func PauseActivity(_ context.Context, roomID string) error {
-	_, err := http.Get(adonisEndpoint + "/temporal/pause/" + url.QueryEscape(roomID))
+	url := adonisEndpoint + "/temporal/pause/" + url.QueryEscape(roomID)
+	fmt.Println(url)
```
Devessier, line 28: We could remove the calls to `fmt.Println`.

Devessier (changes requested):
This is perfect. We now have a pause and play mechanism that works well!

Devessier (approved):
Perfect! We can merge.

PR #36, opened 2021-06-14
Connect Music Player to hardware back button

### LINKED ISSUE #28: Fix hardware previous button on the trackPlayer
After loading a track in the player and clicking on it to go the the MusicTrackPlayerScreen by hitting the previous hardware button react navigationd do not redirect to the previous page. Because we're emulating a kind of redirection that is a real one

Devessier:
When the Music Player is in full screen mode and the hardware back button is pressed, we dismiss the Music Player.
We used [@react-native-community/hooks](https://github.com/react-native-community/hooks).

Closes #28


prastoin (approved):
Won't going back on this one !

PR #40, opened 2021-06-15
Set up postgresql

prastoin:


Devessier (approved):
Very nice!
Now we have a back-end with superpowers!
Lucid and AdonisJS migrations seem to be really delightful 👀 .

PR #45, opened 2021-06-16
Add tracks list to rooms

### LINKED ISSUE #44: Add a tracks list to rooms
Currently rooms can only play a single hard coded song. It's time to upgrade them.

We want a room to have a list of songs to play. The objective of this development step is not to let the user populate this list. We will have a hard coded list. The truly objective is to have users listening to the same song at the same time, and to have songs from the list be played in sequence.

Let's imagine what user flow we want at the end:
1. Daniela, a user of our application, searches the song _L'italien_ of Serge Regiani. She finds it and clicks on it.
2. This action will create a room. The first track of this room will be _L'italien_ of Serge Regiani. _The song is played instantly._
3. When the room have been created, the tracks list have been populated with some hard coded songs, plus the one Daniela selected.
4. Lambert Wilson, another user of our application, joins the room created by Daniela. He will instantly listen to the song currently played in the room. Basically it will be _L'italien_ if Lambert is fast enough.
5. Daniela and Lambert can both control the room and decide to stop the song. _This is currently implemented but we need to ensure we do not break it._
6. When the first song of the room ends, both users will listen the next one automatically.
7. When there are no more tracks to play, the last track remains apparent controls do nothing.

## What we need to do

The behaviour we expect will require some developments.

- [x] Temporal must own the list of the tracks that need to be played. Actually we can just store ids on Temporal.
- [x] When we create a room by clicking on a song, this song must be transmitted to Temporal so that it becomes the first item of the tracks list. **We transmit an array of songs.** It will be easier if we want to start a room with several tracks. And guess what, we planned that. See this article of Swyx on [preemptive pluralization](https://www.swyx.io/preemptive-pluralization/).
- [x] The front-end needs to have some information about the songs, at least the title and the author. We can use [Videos#list](https://developers.google.com/youtube/v3/docs/videos/list) call from YouTube API to get them. It would be the responsibility of the back-end to return these information to the front-end.
- [x] Temporal would have a timer to know when the current video ends, but it would not dictate to clients where in the video they must be. The workflow would start a timer when the song is first played, then it would cancel it when a user pauses it and it would recreate a timer once the song is played again. Clients position in a track would be independent from the workflow: the workflow tells what song to play and reacts to songs end, clients always obey to the workflow but they handle their cursor position in the video themselves. The benefit is that we do not have to perform `seek` operations on every tick. Clients will have to deal with network deficiencies and it would be a pain to control the position in the video of every client. The counterpart is that, as we do not want to loose synchronization between clients about which track is played, if a client is late and the next song has to be played, the end of the song would be cut for this client so that she listens directly to the next track.
- [x] Clients must play the song the back-end wants.
- [x] Clients can click on _next song_ button and the workflow must react.

**Technically we have to do the following things:**

- [x] Have a slice of strings in the workflow that represents the tracks list
- [x] When a song is clicked in the search results, we must send the id of the song to the back-end to create the room if this track filled by default in the track list
- [x] The room must be created with a tracks list prefilled with this song, and some other hard coded songs for now (will have to be coded with _suggestion_ feature)
- [x] Send information about the tracks list to users that join the room
- [x] We must broadcast to clients information about the tracks list: which is the current track, what does tracks list contain
- [x] We must create a timer in the workflow when a song is launched
- [x] The timer needs to be cancelled when the track is paused and recreated when the track is resumed
- [x] When the timer ends, we need to launch the next song, if any, by recreating a timer and sending to all clients which is the new track to play
- [x] #56
- [x] #55
- [x] #54 
- [x] A client that joins a room already playing a track needs to know where to start the track
- [x] #63
- [x] #62
- [x] #84

Devessier:
I am working on this feature.


Devessier:
We refactored the code of the MTV workflow to use a state machine.
We still need to migrate the JOIN action inside the state machine.
We also have to return some information from the state machine context in the state query. We should see if the workflow should continue to own a `state`, or if we should migrate it inside the context of the state machine.

Devessier:
## Check up

In the workflow we implemented the fetching of the first tracks information from YouTube API.
These information are broadcast to the user that created the room and to those who will join it later.
We had to replace the callback the `CREATE_ROOM` Socket.IO event expected to receive by another event, `CREATE_ROOM_CALLBACK`. It was necessary to ensure the acknowledgement of the creation could not be broken by a server shutdown. We had to do the same thing for the `JOIN_ROOM`, with a new event `JOIN_ROOM_CALLBACK`.

prastoin (approved):
Great job ! Temporal has been improve by far with brainy. Loving how it's keeping the xstate style along the front and temporal pkgs
Now last sprint, we need to connect everything together to see the magic happen

PR #59, opened 2021-07-07
Move initial tracks fetching into the stateMachine temporal

prastoin:


packages/temporal/workflows/mtv.go
```diff
@@ -148,9 +148,66 @@ func MtvRoomWorkflow(ctx workflow.Context, params shared.MtvRoomParameters) erro
...
+												internalState.TracksIDsList = internalState.TracksIDsList[1:]
+											}
+										}
+
+										return nil
+									},
+								),
+							},
+						},
+					},
```
Devessier, line 207: As there is only one possible transition for this event, we can simplify the code. It absolutely does not change the behaviour:

```suggestion
					MtvRoomInitialTracksFetched: brainy.Transition{
						Target: MtvRoomPausedState,

						Actions: brainy.Actions{
							brainy.ActionFn(
								func(c brainy.Context, e brainy.Event) error {

								event := e.(MtvRoomInitialTracksFetchedEvent)
								internalState.Tracks = event.Tracks

									if tracksCount := len(event.Tracks); tracksCount > 0 {
										currentTrack := internalState.Tracks[0]
										internalState.CurrentTrack = currentTrack

										if tracksCount == 1 {
											internalState.Tracks = []shared.TrackMetadata{}
											internalState.TracksIDsList = []string{}
										} else {
											internalState.Tracks = internalState.Tracks[1:]
											internalState.TracksIDsList = internalState.TracksIDsList[1:]
										}
									}

									return nil
								},
							),
						},
					},
```

Devessier (changes requested):
We could simplify a transition you added to the state machine. This is not crucial at all but it can conveys the meaning that only what transition can happen there.

packages/temporal/workflows/mtv.go
prastoin: Fixed !

Devessier (approved):
It's perfect! The state machine is really clean.
We can merge.

PR #60, opened 2021-07-08
Handle user joining in the state machine

Devessier:
We moved the code related to adding users in the room inside the state machine.

The `Export` method of `MtvRoomInternalState` type used to call the `Current` method of the state machine to gets its current state, and know if the room was playing. It caused deadlock when we tried to `Export` the internal state from an action of the state machine, as `Current` uses the lock of the state machine to prevent concurrent accesses. This security is unnecessary inside a workflow as all the code is guaranteed to run sequentially and on a single goroutine. Therefore we could replace the use of `Current` by its unsafe equivalent, `UnsafeCurrent`, which does not use the state machine lock.

Closes #60

### DIFF
--- a/packages/temporal/activities/mtv_controls.go
+++ b/packages/temporal/activities/mtv_controls.go
@@ -45,8 +45,18 @@ func CreationAcknowledgementActivity(_ context.Context, state shared.MtvRoomExpo
 	return err
 }
 
-func JoinActivity(ctx context.Context, state shared.MtvRoomExposedState) error {
-	marshaledBody, err := json.Marshal(state)
+type mtvJoinCallbackRequestBody struct {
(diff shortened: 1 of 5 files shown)


prastoin (changes requested):
Cool ! no activities called outside the state machine :).
One last thing, you couldn't know mb could you please send back to Adonis in the joinActivity req body the joining userID. It will be use full to persist the operation in the adonis instance

Then Temporal JoinActivity request body should looks like:

```
{
    state: {
         roomID,
         users,
         tracks,
         ....   //containing the ControlState json format   
    },
    joiningUserID: 'the-user-uuid'
}
```

prastoin (approved):
Great ! thanks

PR #65, opened 2021-07-13
Forward go to next track action from the client to the mtv workflow and play the next track if applicable

### LINKED ISSUE #62: Clients can request to go to the next track
When the user clicks on the _next_ button, the socket.io client must emit an event to the back-end. Be this event be called `GO_TO_NEXT_TRACK`.

The server should forward this event to Temporal by making an HTTP request. It should retrieve the workflow id and the run id of the Temporal workflow instance from `MtvRoom` Lucid model.

The Go REST API should expose a new endpoint that takes a workflow id and a run id in the body. This endpoint should send a signal to the correct Mtv Room Workflow instance. Be this signal be called `go-to-next-track`.
When the workflow receives the signal, it should send an event to the state machine. Be this event be called `GO_TO_NEXT_TRACK`.
When the state machine receives this event, if should assign the first item of `internalState.Tracks` slice to `internalState.CurrentTrack`, if there are items in `internalState.Tracks`, then it goes to `MtvRoomPlayingState` state. Otherwise it does nothing.

## Tasks

- [x] Emit `GO_TO_NEXT_TRACK` from client to server
- [x] Forward this event to Temporal from the server and verify it
- [x] Go to the next track in the workflow state machine
- [x] Write Temporal tests

Devessier:
When the go to next track button is pressed on the client, we send to the server a message through websockets. When the server receives it, it makes an HTTP call to the Temporal REST API. When the Temporal REST API receives the request, it sends a signal to the workflow of the mtv room.
When the mtv workflow receives the go to next track signal, it forwards the event to the state machine. Then, the state machine will transition to the playing state with the current track having been set to the next one, only if there are tracks to play.
These three parts have been tested.

Closes #62

### DIFF
--- a/packages/client/__tests__/ForcedDisconnectionPrintAlert.test.tsx
+++ b/packages/client/__tests__/ForcedDisconnectionPrintAlert.test.tsx
@@ -10,24 +10,8 @@ function noop() {
     return undefined;
 }
 
-function waitForTimeout(ms: number): Promise<void> {
-    return new Promise((resolve) => {
-        setTimeout(() => {
-            resolve();
-        }, ms);
-    });
-}
-
 test(`On FORCED_DISCONNECTION it should displays the alert modal and dismiss it when clicking on dismiss button`, async () => {
-    const {
-        getByText,
-        getByPlaceholderText,
-        getAllByText,
-        findByText,
-        getByTestId,
-        findByA11yState,
-        debug,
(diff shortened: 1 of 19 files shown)


prastoin (approved):
Well done for the first part of the implemtation.
Nice to scope PR.
See you soon for the server acknowledgement :)

PR #66, opened 2021-07-15
Create a wrapper component to use different youtube client library on native and web paltforms

### LINKED ISSUE #39: Playing videos on the Web fails
An error occurs when the music player is shown in full screen.
The Youtube iframe is visible but an error occurs when we call `getDuration` on the Youtube player reference.

CleanShot 2021-06-15 at 10 06 56@2x

Devessier:
On native platforms we must use `react-native-youtube-iframe` library and on the Web we use `react-youtube`.

Thanks to React Native [platform specific code](https://reactnative.dev/docs/platform-specific-code), we load the correct library only when needed.

We used [Devessier/react-native-web-youtube-player](https://github.com/Devessier/react-native-web-youtube-player) as a playground.

Closes #39


prastoin (approved):
Super cool, a thorny issue we can get rid off thanks to you
lets go

PR #68, opened 2021-07-16
Fix broken search bar animations on Web

### LINKED ISSUE #67: Make search screen animations work on the Web
Animations in the search screen do not work on the Web. The result is quite strange.

CleanShot 2021-07-16 at 10 55 37

Devessier:
Animations of the search bar were broken on Web because [`sx` function from dripsy](https://github.com/nandorojo/dripsy#headless-dripsy-with-usesx) does not return an object with resolved values on the Web as it does on native platforms. The reason is that on the Web _react-native-web_ uses CSS Modules for styling. On the Web `sx` returns the id of the created CSS Module instead of an object with the styles.

As a consequence, we had to find a way to, nonetheless, get the values we wanted:

- We created an exported constant, `GLOBAL_THEME_CONSTANTS`, that contains some theme constants that do not differ between light and dark themes. This constant allows type checking without trying to change dripsy internal types.
- For values that are resolved at runtime, depending on the current theme, we use `useDripsyTheme` that returns the current theme. We can extract the property we want, here `greyLighter`, but we loose type safety.

Closes #67

### DIFF
--- a/packages/client/components/kit/AppScreenHeaderSearchBar.tsx
+++ b/packages/client/components/kit/AppScreenHeaderSearchBar.tsx
@@ -1,5 +1,5 @@
 import { useMachine } from '@xstate/react';
-import React, { useEffect, useRef } from 'react';
+import React, { useRef } from 'react';
 import { TextInput as RNTextInput, TouchableOpacity } from 'react-native';
 import { createMachine } from 'xstate';
 import { View as MotiView } from 'moti';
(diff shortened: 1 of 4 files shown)


prastoin (approved):
This a good thing to keep checking for cross platform improvement along the whole dev we've done. This is still a main point of this project
Congratulations
Ready to merge !

PR #72, opened 2021-07-16
Fix retrieve context

### LINKED ISSUE #73: Fix retrieve context
The retrieve context is not working on mobile because the socket listeners aren't init before the socker server event 'connection' that we're using to sync the new coming session mtv room context. Atm on browser it seems to be working, but it's quite random.

The solution is to stop using the server socket connection event to synchronize the session. But thanks to an invoke in the state machine first initial state, where we emit as a bottle in the sea a GET_CONTEXT event.
If it occurs in the back that the socket is related to an existing mtvroom then the server will emit to this socket a RETRIEVE_CONTEXT event that will be catch by the state machine at its root level.
We do not use callback as it make the state machine goes crazy and infinite looping on it's initialization

prastoin:
close #73

### DIFF
--- a/packages/client/machines/appMusicPlayerMachine.ts
+++ b/packages/client/machines/appMusicPlayerMachine.ts
@@ -150,6 +150,15 @@ export const createAppMusicPlayerMachine = ({
 
             states: {
                 waitingJoiningRoom: {
+                    invoke: {
+                        src: (_context) => () => {
+                            /**
+                             * Looking for other sessions context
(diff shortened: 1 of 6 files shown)


prastoin:
close #73

Devessier:
Great. The bug was difficult to spot.
We can merge.

Devessier (commented):
Great. The bug was difficult to spot.
We can merge.

Devessier (approved):
We can merge.

PR #74, opened 2021-07-19
Fix two alive sessions sync create callback

### LINKED ISSUE #71: Sync two parallel sessions on createRoom
Start two sessions.
Create room from one, the other one should display the created room too

Solution:
Bring down to low level the CREATE_ROOM_CALLBACK listener in the state machine to force the transition even if the specific state machine is not the CREATE_ROOM emitter

prastoin:
Be aware that you'll have headache after reviewing this PR
close #71


Devessier (approved):
It's perfect. State machines ease so much things like that.
We can merge.

PR #76, opened 2021-07-19
Fix initial track duration passed to the mtv workflow timer

### LINKED ISSUE #70: With two sessions when we play then pause on one, then refresh the other, it goes to the next track
We need to check why and fix this.
Maybe currentTrack is lost in the nature in temporal

Reproduce the bug:
1. Start one session create a room on it then hit play button.
2. Start another session, it should be synced to the previous one, but here is the bug, current track is not synced with the first session. It appears to be the next track from the waiting list and also the room is not playing but the song looks to be paused, it shouldn't

Devessier:
The timer of the first initial track was not launched with correct parameters. It used to receive an empty duration when we were expecting to get the duration of the first track. It has been fixed.

We also told to nodemon to track all go files. There was an issue with the command of the worker, that nodemon could not parse correctly and then could not determine which files to track on its own.

Closes #70

### DIFF
--- a/packages/temporal/package.json
+++ b/packages/temporal/package.json
@@ -7,8 +7,8 @@
         "yarn": "1.22.10"
     },
     "scripts": {
-        "dev": "nodemon --signal SIGTERM --exec env PORT=3000 ADONIS_ENDPOINT='http://localhost:3333' go run api/main.go",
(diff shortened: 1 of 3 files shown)


prastoin (approved):
Really cool that you found out this issue from an really surprising behaviour.
Lets go mergin !

PR #77, opened 2021-07-19
Improve first create room callback

### LINKED ISSUE #75: Improve first create room callback global sync between user sessions
Atm only the emitting CREATE_ROOM event device receives the callback as it's in the socket event payload.
It means the room_is not_ready view will be display only in this specific device.
We could refactor that by emitting a server socket event that notifies every user's sessions that a room is in creation.

prastoin:
close #75


Devessier (approved):
It's great to continue to use two different events to do a RPC.
We can merge.

PR #78, opened 2021-07-19
Fix remote join timeout issue

### LINKED ISSUE #69: RemoteJoin timeout blockin process
On join nor create socket event, we need to connect every involved user's devices to the room. If the a device database is corrupted the remote join won't achieve.
It throw the error remoteJoin timeout.
We need to silent error this throw

prastoin:
close #69


Devessier (approved):
Great! You wrote our first real unit test and it is totally justified. It's a good sign that it did not occur before.
We can merge!

PR #79, opened 2021-07-19
Broadcast go to next track to clients

### LINKED ISSUE #63: Clients need to play the video the server wants
When the current track changes, the Mtv Room Workflow should broadcast the information to all the clients in the room.

It should make a request to the server, with the room id in the body as well as the workflow state.

The server should expose an endpoint to handle this request. The controller of this route should confirm that there is such a room in the application, then it should send an event inside the socket.io room of the mtv room with the content of the tracks list and the current track. Be this event be called `FORCE_UPDATE_CURRENT_TRACK`.

The client should listen to this event from the `appMusicPlayerStateMachine`. When we receive this event, we should switch to the new current track and play it directly.

Devessier:
As described in the related issue, we found that the easiest way to implement go to next track broadcasting was to send the state of the mtv workflow with the `ACTION_PLAY_CALLBACK` event.

When a client receives this event, the state of the workflow is merged with the context of the appMusicPlayer state machine.

Closes #63


Devessier:
Currently the go to next track information is broadcasted to clients and the next track is put in the player. However, it is not automatically played.
We need to wait for the player to load the track and then play the track.

Devessier:
The native player now automatically plays next tracks.

**There is a bug on the Web that cause videos to do not play.**

Devessier:
The next tracks did not play because the `onReady` function was not called for subsequent tracks. We fixed that by listening to player state changes, and when the player reaches the `UNSTARTED` state we call `onReady` function.

It appeared that when we tried to get the duration of a next track, the player returned `0`. We tried to find a way to wait for the player to be really in a ready state to get a non-zero duration, but we did not succeed. _We deleted the duration fetching for native and web, and now use the duration returned by YouTube API._

Devessier:
There is an issue when we reach the end of the last track.
When we try to play the track, Temporal calls the PlayActivity, which tells clients to play the track.
We would expect the mtv workflow to do not call the PlayActivity if we reached the end of the current track and if there are no more tracks to play.

CleanShot 2021-07-21 at 15 44 27

Devessier:
We fixed the bug and added tests to prevent a regression.

Devessier:
I think the PR is ready to be merged.

prastoin (approved):
Really nice implementation of the go to next track feature inside the play_callback.
Lets go mergin !

PR #81, opened 2021-07-20
Refactor join room funnel

### LINKED ISSUE #80: Refactor Join a room funnel
As temporal now sends a use local activity to notify about the join success operation we need to adapt the funnel.
Then in Adonis we need to emit and make the socket join the room only on the temporal request.

Improvement:
At the moment when a user joins a room, he won't retrieve the currentTrack timecode. That means that if the song for first user in a room has been played for half the total duration, any newcoming user will have to start the current track again.
In this way in temporal we have to find  way to get the current timer value without canceling/disturbing this one. then we will send the elapsed time in the current track.

Todo

Client:
 - [x] Nothing

Adonis:
 - [x] Move the logic in the temporal response
 - [x] Update test
 
 Temporal:
 - [x] Find a way to extract elasped time in timer or a way to calculate it

prastoin:
close #80


packages/temporal/workflows/mtv_test.go
```diff
@@ -382,32 +362,33 @@ func (s *UnitTestSuite) Test_JoinCreatedRoom() {
...
+	s.env.ExecuteWorkflow(MtvRoomWorkflow, params)
 
 	s.True(s.env.IsWorkflowCompleted())
 	err := s.env.GetWorkflowError()
 	s.ErrorIs(err, workflow.ErrDeadlineExceeded, "The workflow ran on an infinite loop")
 }
 
 // Test_GoToNextTrack scenario:
-//
+
```
Devessier, line 373: We should keep the empty comments. If you hover the function `Test_GoToNextTrack` you will see that its description only begins from the fifth point. Empty comments are intended to allow line breaks.

prastoin: Done !

prastoin:
We need to fix the web player issue on component mounting and state.Playing true.
About to write an issue monday 09

packages/temporal/go.mod
```diff
@@ -3,15 +3,14 @@ module github.com/AdonisEnProvence/MusicRoom
...
-	github.com/Devessier/brainy v0.0.10
+	github.com/Devessier/brainy v0.0.11
 	github.com/bojanz/httpx v0.0.0-20201111190843-d1cf01c49b2e
 	github.com/bxcodec/faker/v3 v3.6.0
 	github.com/go-playground/universal-translator v0.17.0 // indirect
 	github.com/go-playground/validator v9.31.0+incompatible
 	github.com/golang/mock v1.6.0 // indirect
 	github.com/gorilla/handlers v1.5.1
 	github.com/gorilla/mux v1.8.0
-	github.com/joho/godotenv v1.3.0 // indirect
```
prastoin, line 14: We should manage this issue for next new repos. With for example a js lib dotenv

Devessier (approved):
Very heavy work. Great job man!
We can merge and go on.

PR #83, opened 2021-07-22
Hide properly full screen player when not on full screen mode

### LINKED ISSUE #82: Layout issue with the music player on the web
On the Web, when the full screen music player is shown, that is once the user is in a room, it is taking all the height of the screen, event if it is not in full screen mode and should be hidden outside the bottom of the screen.

image

Devessier:
The player was supposed to be hidden, but it was taking all its own height.
We had to position it absolutely outside the right bottom of the screen.

**There is currently an issue with the Web player: when it's size changes, the video is stopped. This issue will have to be fixed in the future because the player is also stopped when the player is reduced or maximized.**

Closes #82


prastoin (approved):
Cool fix
Shall we static set the iframe dimensions for the web platform as it has issues rn
We will create a specific issue about it

PR #85, opened 2021-08-05
Design full screen player

### LINKED ISSUE #84: Design the player
It's time to improve the design of the music player.

We need to do the following things:

- [x] Reduce the size of the controls. Currently they take too much space
- [x] Create two subviews under the player:
  - [x] Tracks
  - [x] Chat
- [x] Make the player fixed
- [x] Create a noop button to suggest a song
- [x] Make the player look nice on Web
- [x] Prevent the Web player from stopping when put in background
- [x] Fix issues with player unmounts

image

Devessier:
We designed the tracks list. We put the Suggest Track floating button at the right bottom of the tracks list.

We fixed several bugs with the web player. It should work properly now: we can minimize it, resize the window.

Closes #84


prastoin (approved):
Great work, keeping in mind the cross platform goal is really important and you dealt perfectly with it here.
Also the mtvRoom is really becoming concrete with the tracks cards.
Lets go mergin !

Issue #88, opened 2021-08-07
As a user choose a device to play music on


prastoin:
### Choose a device to play music on
As we can now relate devices to a user. We should now be able to implement the choose a device to play the music on feature.

The user should be able via the music player full screen to see his alive-sessions related ( connected ) devices.
And then be able to switch the playing music device.
This features is all about: 
Allow the user to choose a device to play the music on.
To be able to list connected user's devices

Technically we need to do:

Client:
---
 - [x] Put the musicPlayer volume to 0 if the device is not the emitting one
 - [x] At the contrary set the volume to 100 if the device is the emitting one.
 
Both of the points above are computes by the client through a new listened socket event `"ACTION_CHANGE_EMITTING_DEVICE_CALLBACK"`.
As always the payload will be a shallow copy of the `MtvWorkflowState` at time T.
It'll be then merged in the `appMusicPlayerStateMachine` via the `assignMergeNewState`.
It involves to update the `exposedMtvWorkflowState` to persist user scoped information, such as the emitting device uid.
 - [x] List the users's currently alive sessions ( connected ) devices, and allow him to pick one as emitting
 It involves to update the `appMusicPlayerContext` to persist user connected devices, such as context.devices ( string[ ] )
- [x] Create a context provider around the user informations
- [x] Create a state machine that will be useful later for the auth, but where for the moment we will listen to `CONNECTED_DEVICES_UPDATE` socket event

This feature will be the first step in the Authentication/User scoped development.
We would like to create a context provider that will manage the user related information, such as here specifically the user connected devices listener etc
 
Server nor Types:
---
- [x] Create and handle `"CHANGE_EMITTING_DEVICE"` socket event, containing a device uid in the payload. 
Check if the given `deviceUid` is well emitting user related
It should then persist the operation to the user corresponding mtv room workflow
- [x] Create and handle the new temporal Http callback request corresponding to the "CHANGE_EMITTING_DEVICE" acknowledgment and send back a socket 
"ACTION_CHANGE_EMITTING_DEVICE_CALLBACK" event to the user socket instance.
- [x] Add the user emitting device uid to the CREATE_WORKFLOW and JOIN_WORKFLOW payloads and persist them in the temporal `internalState`. Default emitting device uid will be the socket event emitter.
- [x] test above features

[...]

PR #91, opened 2021-08-10
List user devices to front

prastoin:
This Pr is related to the issue #88


Devessier (approved):
Great job. We will have a system that works better then Spotify one.
We can merge.

PR #97, opened 2021-08-24
Create suggest song modal

Devessier:
We created the modal that allows a user to select a track to suggest.
We tested the behaviour.
We still need to implement the back-end logic and connect it to the front-end.


prastoin (approved):
First step in the suggest track, great !
Well done
Ready to merge

PR #98, opened 2021-08-24
Load environment variables correctly in temporal package

### LINKED ISSUE #90: Use `dotenv` NPM library in Temporal scripts
We were using `godotenv` for one script, but it has been removed from the list of dependencies by `go mod tidy` as it is not used in the code.
We should use [dotenv](https://github.com/motdotla/dotenv#readme) as it will not lead to similar issues.

Devessier:
We did not use dotenv as it only works with JavaScript programs. Instead we used [env-cmd](https://github.com/toddbluhm/env-cmd) that tries to load `.env` files and launches the provided command with the extracted environment variables.

Closes #90

### DIFF
--- a/packages/temporal/.env.example
+++ b/packages/temporal/.env.example
@@ -1,0 +1,3 @@
+GOOGLE_API_KEY=""
+PORT="3000"
+ADONIS_ENDPOINT="http://localhost:3333"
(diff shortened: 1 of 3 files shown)


prastoin (approved):
Great it needed to be done
lets go !

PR #105, opened 2021-08-26
Forward suggestions from client to temporal

### LINKED ISSUE #87: Tracks suggestion
Now we have a tracks list that works. Tracks can be played in sequence, each device connected to a room receives the updates in real time, tracks can be paused and resumes as many time the user wants and we have a so much better implementation of tracks timers in the Temporal Workflow.
The next step is to implement tracks suggestion. We no longer want to have a fixed tracks list, we want to populate it through the user interface.

The first step could be to have a _suggested tracks list_ that would be shown under the tracks list on the player. The user would suggest songs through a well defined flow:

1. The user clicks on _Suggest a Track_ button
2. A tracks search screen is shown to the user. _We need to determine how we implement it technically._ Do we hide the full screen and redirect the user to a screen with a search bar? Do we show a modal? Spotify uses a modal to show a tracks selector when the user wants to add tracks to a playlist (see below).
3. The user can search for a song. If the user clicks on the _Add track_ button, the song must be added to the suggestion list
4. The user can also add songs from the main Search Track screen, linked into the bottom bar. See image below.

## The Workflow

In the Temporal Workflow, we should have a slice that owns all the suggested tracks. This slice would be of type `[]shared.TrackMetadata`.
When we receive, through a signal, that a user suggested a track, we should compare the id the signal contains with all the scheduled tracks and all the suggested tracks. We do not allow duplicates. However, if the suggested track has already been played, we allow it to be suggested again.
Once we verified that the suggested track is not a duplicate, we should fetch its information from YouTube API and assign it to the suggested tracks slice.

When a suggested track is accepted, we should call an activity that broadcast the information to all the clients. To ease the flow we can send all the workflow state, as we did previously.

## UX

_What do we show to the user while we verify that the track she suggested can be added? Does she stay on the search track screen, with a loading spinner?_

### Front-end

- [x] Create search tracks modal
- [x] Show search track modal when _Suggest a Track_ button is pressed
- [x] Allow to quit the modal
- [x] When a song is pressed, close the modal and go back to the full screen player
- [x] Send the selected track to suggest with an event to the back-end
- [x] Wait for the back-end acknowledgement through a WS event
- [x] Show a toast when we received the response
- [x] Wait for an event that is sent by the back-end when a track is suggest ; merge the workflow state with the machine context

### Back-end

- [x] Wait for events sent by the client with a track to suggest
- [x] Broadcast the request to Temporal
- [x] Wait for Temporal acknowledgement of the suggestion
- [x] Broadcast the response of Temporal to the client

### Temporal

- [x] Wait for requests sent by the back-end with the track to suggest
- [x] Send it with a signal to the workflow
- [x] Create a list of suggest tracks
- [x] Append the suggested track to the list if there are no duplicates
- [x] If the track could be suggested, broadcast the information to the back-end by sending the workflow state

### Spotify tracks list adding modal

image

### Options on searched tracks

image

### LINKED ISSUE #102: Notify suggesting user about the success/fail operation
As we've discuss during peer review, we do not want a user to be able to spam suggesting track operation.
In this way we need to persist the status of the current user suggest track from temporal to the client to update the state machine.
?? Maybe should we also have an adonis verification about the user's ability to make a suggest ??

Comment findable in packages/temporal/workflows/mtv.go line 670:
```
// The suggesting track user needs to receive the operation acknowledgement
// to be able to resuggest a new track
// In this way near this area we should send a acknowledgement activity
// that will contain a suggesting user related state
// Then we will send to every suggesting user a specific socket event
```

Devessier:
Suggestions are now fully implemented.
The user will see an activity indicator while we are waiting for temporal acknowledgement. Once the acknowledgement came, we trigger a toast.

Everything works on native platforms and on the Web.

Users are informed that their request has been taken into account.
Closes #102

It marks the end of tracks suggestion step.
Closes #87

### DIFF
--- a/packages/client/navigation/RootNavigation.tsx
+++ b/packages/client/navigation/RootNavigation.tsx
@@ -14,17 +14,6 @@ export const isReadyRef: React.MutableRefObject<boolean | null> =
 
 export const navigationRef = React.createRef<NavigationContainerRef>();
 
-export function useGlobalNavigationRef(): NavigationContainerRef | null {
-    const isReady = isReadyRef.current;
-    const navigation = navigationRef.current;
-
-    if (isReady === true && navigation !== null) {
-        return navigation;
-    }
-
-    return null;
-}
-
 // eslint-disable-next-line
 export function navigateFromRef<Route extends NavigateFromRefRoutes>(
     name: NavigateFromRefRoutes,
--- a/packages/client/navigation/RootNavigation.tsx
+++ b/packages/client/navigation/RootNavigation.tsx
@@ -14,6 +14,17 @@ export const isReadyRef: React.MutableRefObject<boolean | null> =
 
 export const navigationRef = React.createRef<NavigationContainerRef>();
 
+export function useGlobalNavigationRef(): NavigationContainerRef | null {
+    const isReady = isReadyRef.current;
+    const navigation = navigationRef.current;
+
+    if (isReady === true && navigation !== null) {
+        return navigation;
+    }
+
+    return null;
+}
+
 // eslint-disable-next-line
 export function navigateFromRef<Route extends NavigateFromRefRoutes>(
     name: NavigateFromRefRoutes,
--- a/packages/server/start/socket.ts
+++ b/packages/server/start/socket.ts
@@ -323,10 +323,13 @@ Ws.io.on('connection', async (socket) => {
 
         socket.on('SUGGEST_TRACKS', async ({ tracksToSuggest }) => {
             try {
-                const { mtvRoomID } =
-                    await SocketLifecycle.getSocketConnectionCredentials(
-                        socket,
-                    );
+                const {
+                    mtvRoomID,
+                    user: { uuid: userID },
+                    deviceID,
+                } = await SocketLifecycle.getSocketConnectionCredentials(
+                    socket,
+                );
                 if (mtvRoomID === undefined) {
                     throw new Error(
                         'Can not go to the next track, the user is not listening to a mtv room',
@@ -336,6 +339,8 @@ Ws.io.on('connection', async (socket) => {
                 await MtvRoomsWsController.onTracksSuggestion({
                     roomID: mtvRoomID,
                     tracksToSuggest,
+                    userID,
+                    deviceID,
                 });
             } catch (err) {
                 console.error(err);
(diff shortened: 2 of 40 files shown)


packages/types/src/mtv-room-websockets.ts
```diff
@@ -67,4 +72,6 @@ export interface MtvRoomServerToClientEvents {
     FORCED_DISCONNECTION: () => void;
     JOIN_ROOM_CALLBACK: (state: MtvWorkflowState) => void;
     CHANGE_EMITTING_DEVICE_CALLBACK: (state: MtvWorkflowState) => void;
+    SUGGEST_TRACKS_CALLBACK: (state: MtvWorkflowState) => void;
```
prastoin, line 75: Shall we rename the SUGGEST_TRACKS_CALLBACK to something like SUGGEST_TRACKS_LIST_UPDATE as it's client socket event related.
Also ACKNOWLEDGE_TRACKS_SUGGESTION might be more the SUGGEST_TRACKS_CALLBACK as it's related to the client socket event SUGGEST_TRACKS

packages/server/start/socket.ts
```diff
@@ -320,6 +320,32 @@ Ws.io.on('connection', async (socket) => {
...
+                const {
+                    mtvRoomID,
+                    user: { uuid: userID },
+                    deviceID,
+                } = await SocketLifecycle.getSocketConnectionCredentials(
+                    socket,
+                );
+                if (mtvRoomID === undefined) {
+                    throw new Error(
+                        'Can not go to the next track, the user is not listening to a mtv room',
```
prastoin, line 335: Change error label

packages/client/navigation/RootNavigation.tsx
```diff
@@ -14,6 +14,17 @@ export const isReadyRef: React.MutableRefObject<boolean | null> =
 
 export const navigationRef = React.createRef<NavigationContainerRef>();
 
+export function useGlobalNavigationRef(): NavigationContainerRef | null {
```
prastoin, line 17: to rm

packages/types/src/mtv-room-websockets.ts
Devessier: Done

packages/server/start/socket.ts
Devessier: Done

packages/client/navigation/RootNavigation.tsx
Devessier: Done

prastoin (approved):
Great job !
Nice work around the suggest track feature.
Ready to merge
Strength for the following suggesting q feature

PR #111, opened 2021-08-31
Allow several simultaneous tracks suggestions

### LINKED ISSUE #104: Prevent race conditions in tracks suggestion
When tracks are suggested, we overwrite a workflow future with the one returned by tracks information fetching activity.
Once the future is ready, we handle it and save the information we got from YouTube.

The current issue is that if a tracks suggestion occurs while a previous fetching is still pending it will be discarded in favour of the new one.

The problem could be easily solved if we had parallel states in brainy. But we do not want to bet on that feature.
Nonetheless, we should find a way to prevent such race conditions.
Here are some ideas:

## Fill a queue while a fetching is already pending

If fetching is already occurring, do not launch another fetching. Instead, fill a queue with the suggested tracks.
We could then use a [custom future](https://pkg.go.dev/go.temporal.io/sdk/workflow#NewFuture) and [set its value](https://pkg.go.dev/go.temporal.io/sdk@v1.9.0/internal#Settable) to the list of suggested tracks and select it in the infinite loop of the workflow.

Devessier:
We used a simpler method that what we described in the issue.
We have a slice of `workflow.Future` and we fill it each time tracks are suggested.
In the event loop we remove futures from the slice once they have been handled.

Closes #104

### DIFF
--- a/packages/temporal/workflows/mtv.go
+++ b/packages/temporal/workflows/mtv.go
@@ -305,11 +305,11 @@ func MtvRoomWorkflow(ctx workflow.Context, params shared.MtvRoomParameters) erro
 	channel := workflow.GetSignalChannel(ctx, shared.SignalChannelName)
 
 	var (
-		terminated                              = false
-		workflowFatalError                      error
-		timerExpirationFuture                   workflow.Future
-		fetchedInitialTracksFuture              workflow.Future
-		fetchedSuggestedTracksInformationFuture workflow.Future
+		terminated                               = false
+		workflowFatalError                       error
+		timerExpirationFuture                    workflow.Future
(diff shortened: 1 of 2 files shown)


packages/temporal/workflows/mtv.go
```diff
@@ -1010,6 +1011,11 @@ func MtvRoomWorkflow(ctx workflow.Context, params shared.MtvRoomParameters) erro
 	return workflowFatalError
 }
 
+func removeFutureFromSlice(slice []workflow.Future, index int) []workflow.Future {
```
Devessier, line 1014: Inspired by https://stackoverflow.com/a/37335777.

prastoin (approved):
Great solution !
Lets go merging

PR #116, opened 2021-09-03
Add mtv room creation form

Devessier:
## Bugs to fix

- [x] When we go out of the form the application crashes because `useActor` receives an `actorRef` of `undefined`


prastoin (approved):
Wow great work
Incredible use of xstate/tests
State machines to the win
Lets go mergin

PR #119, opened 2021-09-06
Vote for track client

### LINKED ISSUE #110: Vote for a track
As for now a user can suggest a song.
From here we need to allow a user to vote for a song.
Note: a user can vote for several tracks in a row, we do not need to wait the end of the operation
Then add a score to each track, depending on this score order the tracks list from the highest score to the lowest.
The first one, the track with the highest score, will be the next played ( or the first played ).
During the room creation the creator can defined the minimum score needed for a track to be played.
For the moment, until the room creation pull request/dev, this data will be mocked to 2 votes.
In this way a track will be found under two statuses:
1- About to be played, eligible score >= minimum score
2- Suggested tracks, score < minimum score
No matter the track status any tracks can receive votes until he's loaded in the player
When a user has voted for a track he cannot remove his vote.
A user can vote for a track by clicking on the track card, from the track's list in the music player full screen view.

When a user has voted for a song a visual element needs to notify the operation ( e.g change track card background color ).
Note: This involve to persist the user vote between every user devices => to develop

After a user has voted for a song every other room's members need to see their track's list score & order updated.
The suggestedTracks & eligibleTracks arrays order are deterministic 

**_New sockets event:_**
**From Client:**
1.  Vote for a track -> `USER_VOTE_FOR_TRACK`

**From server:**
User related:
1. Vote for a track temporal response -> `USER_VOTE_FOR_TRACK_CALLBACK`
The above socket event includes the user's voted for tracks , inside the `state.userRelatedInformations.`

Global scope:
1. Score has been updated for a track in the list -> `TRACK_LIST_SCORE_UPDATE`
This could really be heavy to send a socket event for every user's vote. Maybe shall we want to perform a
TRACK_LIST_RECURRENT_SCORE_UPDATE. that would be triggered every T seconds by temporal

Client:
- [ ] #122
- [x] #121
- [x] Print track score confronted to the minimumScoreToBePlayed in the tracks card list
- [x]  Modify the suggest user related callback to persist a state copy to merge in the state machine, to notify the state machine that user has voted for the suggested track
- [x] Emit a `USER_VOTE_FOR_TRACK` by pressing on a track card.
- [x] Notify to the user tracks he has already voted for. By pressing on these nothing will occurs.
- [x] In the AppMusicPlayer listen to the `USER_VOTE_FOR_TRACK_CALLBACK` and merge the given state.
- [x] In the AppMusicPlayer listen to the `TRACK_LIST_SCORE_UPDATE` and merge the given state.
- [x] Tests

Server:
- [x] In socket.ts create a listener/handler for `USER_VOTE_FOR_TRACK` and persist the operation to temporal
- [x] Create a route and handler for the temporal response to `VOTE_FOR_TRACK` operation. From there emit a `USER_VOTE_FOR_TRACK_CALLBACK`
- [x] ~~Create a route and handler for the temporal `trackListScoreUpdate` activity. From there emit a `TRACK_LIST_SCORE_UPDATE` to every user in room~~ Handled in the SuggestOrVoteActivity thread
- [x] Tests

Temporal
- [x] Add to the internalStateUser definition an array for user voted track.
- [x] Remove SuggestedTrack array from the internalState, only use Tracks array and differentiate suggested to toBePlayed tracks based on their score, also update tests
- [x] Create a route and a handler for the `userVoteForTrack`
- [x] Handle the `userVoteForTrack` signal in the workflow
- [x] Populate user `votedForTracks` array after `userVoteForTrack` signal then send back two activity one user related `userVoteForTrackCallback` and an other one not user related `trackListUpdate` both given state as parameters.
- [x] After a suggest if the suggested track is already in the mtv room tracks list just increment the track score by one. Still sending back suggest user related activities to fulfill client expectations.
- [x] Initial tracks are counted as creator voted by
- [x] After a suggest the injected suggestedTrack will get score = 1 and user tracksVotedFor will be append from the suggested track. Just to say that a user suggesting a track is also automatically voting for it
- [x] Create a method to sort the tracks list arrays directly in temporal after a score's update, in the vote for track signal management
- [x] When a track is loaded in the player as a current track it must be removed from every TracksVotedFor users's array. The only way we have, without giving our souls the evil, is to loop over every users searching for the track ID that has been loaded.
We've discuss about the idea of storing the source  of truth, and even about iterating with two sources, somewhere else from the user definition.
We conclude this would be the optimal performances place, better than the track itself
- [x] refactor the deepEqual tracks methods using a custom function and not reflrect.DeepEqual
- [x] Cannot suggest the currentTrack
- [x] To notify to every users in the room use as for a suggest than a vote the activity/route
- [x] To do not overload the temporal service send the described above activity using an interval that will be init after a vote or a suggest into vote and die if the interval get the same state between two timeout
- [x] If a track has a score less than the minimumScoreToBePlayed he can't be assign as the next current track. In this way in the method assign next track do not allow a low score track to be loaded.
- [x] With the very initials tracks, if the minimumScoreToBePlayed is higher than 1, does the first one will not be auto played/loaded in the player he will need to become playable such as every other tracks.
- [x] Handle the minimumscore init in the create route
- [x] Make fail the go to next track that achieves to the music player freeze
- [x] When no more track are playable in the list, go to pause state and listen to the tracks score update to auto play if possible the very next playable track
- [x] Tests

prastoin:
unfortunately this Pull request name is quiet outdated. It also concerns temporal features, such as not allow next track to load in the player if he's not ready to be played and some auto play mechanics

close #110

### DIFF
--- a/packages/client/__tests__/SuggestTrack.test.tsx
+++ b/packages/client/__tests__/SuggestTrack.test.tsx
@@ -94,7 +94,7 @@ test(`A user can suggest tracks to play`, async () => {
             });
         });
 
-        serverSocket.emit('SUGGESTED_TRACKS_LIST_UPDATE', initialState);
+        serverSocket.emit('VOTE_OR_SUGGEST_TRACKS_LIST_UPDATE', initialState);
         serverSocket.emit('SUGGEST_TRACKS_CALLBACK');
     });
 
--- a/packages/client/__tests__/VoteForTrack.test.tsx
+++ b/packages/client/__tests__/VoteForTrack.test.tsx
@@ -1,0 +1,142 @@
+import { MtvWorkflowStateWithUserRelatedInformation } from '@musicroom/types';
+import { NavigationContainer } from '@react-navigation/native';
+import { datatype, random } from 'faker';
+import React from 'react';
+import { RootNavigator } from '../navigation';
+import { isReadyRef, navigationRef } from '../navigation/RootNavigation';
+import { serverSocket } from '../services/websockets';
+import { generateTrackMetadata } from '../tests/data';
+import { fireEvent, render, within } from '../tests/tests-utils';
+
+function noop() {
+    return undefined;
+}
+
+test.only(`
+User should go to the musicPlayer into the tracks tab and hit a track card to vote for it
+After the vote has been accepted a checked icon will appears next the track score
+Also the scroll view will be updated to the track card position
+`, async () => {
+    const deviceID = datatype.uuid();
+    const deviceName = random.word();
+    const userID = datatype.uuid();
+
+    const state: MtvWorkflowStateWithUserRelatedInformation = {
+        roomID: datatype.uuid(),
+        name: random.word(),
+        playing: false,
+        usersLength: 1,
+        userRelatedInformation: {
+            emittingDeviceID: deviceID,
+            userID,
+            tracksVotedFor: [],
+        },
+        currentTrack: null,
+        roomCreatorUserID: userID,
+        tracks: [
+            generateTrackMetadata({
+                score: 0,
+            }),
+            generateTrackMetadata({
+                score: 0,
+            }),
+            generateTrackMetadata({
+                score: 0,
+            }),
+            generateTrackMetadata({
+                score: 0,
+            }),
+            generateTrackMetadata({
+                score: 0,
+            }),
+            generateTrackMetadata({
+                score: 0,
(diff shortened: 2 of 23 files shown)


prastoin:
I still need to implem front tests and little improvement listed in the related issue

Devessier (approved):
You did a really great job!
Your updates to the Temporal workflow are clean.
Great UI choices.
We can merge.

PR #123, opened 2021-09-08
Handle suggest track error

### LINKED ISSUE #118: Handle suggest failure
When a suggest fails, because is already in the player or what ever we need to send back the informations to the client to udpate its suggest ability

prastoin:
close #118


Devessier (approved):
Good job, now users can send us shitty data and we handle it correctly.
Spawning activities is clearer when extracted in a separate function. It's great.
We can merge.

PR #129, opened 2021-09-10
Implement real inputs for physical constraints step

### LINKED ISSUE #16: `babel-plugin-inline-dotenv` does not load environment variables declared in a `.env` file
We should see why it does not work.
Currently it is not crucial so we do not fix it.

Devessier:
We replaced basic `TextInput` with enhanced components for place, radius, starts at and ends at physical constraints.

We fixed an old issue with babel plugins.
Closes #16


prastoin (approved):
Really nice work
Strength for the following devs !
Lets go mergin

PR #131, opened 2021-09-14
Connect mtv room creation form with backend

Devessier:


prastoin (approved):
Great work
Client tests are becoming smarter every PR
Lets go mergin

PR #132, opened 2021-09-16
fix(client): decode query as the query needs to be also drop database…

prastoin:
… between each tests


Devessier (approved):
Good catch.
We can merge this fix.

PR #133, opened 2021-09-17
Device position client and temporal

### LINKED ISSUE #126: Geoloc polling and constraints
A user can vote in a geoloc constrained mtvRoom if at least one his device fit the required time zone and position.
The position is not and cannot be a trusted value as it's from the client. We can't do anything about it.
Then we might delegate some of the process to the client.
A device will send to the server every 1 minute if it's position changed from at least DELTA meters it will update it's model in database.
After each position update operation, if the device's owner is in a mtvRoom set with geoloc constraints, will triggers a DoesUserPositionFitTheCurrentMtvRoomPositionConstraint operation.
In this verification we will check if any user related devices from database fit the required position. Either false or true the result will persisted to temporal that will update the userRelatedInformation in a boolean called UserFitPositionConstraint.
Temporal will manage by it's own the time constraint.
By storing in it's state the from to dates after each vote he will check if the time.Now() fits.

Note:
- [x] Add the mode direct/broadcast props in the mtvRoomCreation funnel

Client:
- [x] Create a device state machine OR implem the described devs in the existing user state machine
- [x] ask for geoloc device access device with expo ? or react native
- [x] Create a 1 mins interval that will send device position if the difference between the last positionSaved is > to DELTA  meters
- [x] when joining/creating a room with constraint ask for the device geoloc auth if it's not already the case
- [x] Put a button in the room setting tab where the user will be able to manually authorize geoloc
- [x] Test the fact that we ask for location when a user create or join or retrieve a room with constraints or hit the get my location button
- [x] Tests

Server:
- [x] Add to the device model a lng & lat  props 
- [x] Add to the mtvRoom model a constraint lng & constraint lat but also a hasConstraints boolean props also radius
- [x] from an address ask google maps for lat and lng geocoding using this [lib](https://github.com/googlemaps/google-maps-services-js)
- [x] Handle the DEVICE_POSITION_UPDATE signal, update the given device with the given new values and if needed call DoesUserPositionFitTheCurrentMtvRoomPositionConstraint
- [x] Create a function that will compute if the given user fits it's realted mtvRoom geoloc constraint, persist the result to temporal
- [x] handle updateUserFitsPositionConstraint temporal response and persist to the front usign a global server socket event named `USER_PERMISSIONS_UPDATE`
- [x] Tests

Temporal:
- [x] Add to the userRelatedInformation the bool prop UserFitPositionConstraint 
- [x] Create route & signal for the update fitThePositionConstraint
- [x] add to the internal state the start and end dates constraints. Using time lib check if time.Now fits the given scope
- [x] Send back an request to temporal about the userFitsPosition success operation
- [x] Update temporal tests to the new internalState definition
- [x] ?? Tests ??

prastoin:
close #126


Devessier (approved):
Great massive work. Good job!
We can merge.

PR #138, opened 2021-09-22
Allow NPM dependencies to be cached and reused

Devessier:


prastoin (approved):
Great improvement !
Nice to have after shipping the new tests in the client

PR #143, opened 2021-09-24
Rerendering issue client

### LINKED ISSUE #122: When you click on a track card that needs scroll to be in the view, the scroll is reset afterwards make something to prevent that, or maybe scroll into the freshly voted for track scroll position

Devessier:
In this PR we fix the scrolling issue that occurred with tracks list FlatList.

Closes #122


Devessier:
One thing to notice is that FlatList component expects to be always present on the screen so that it can optimise items rendering.
We might want to always render each tab component, but hiding them conditionally.

prastoin (approved):
Nice improvement on the client semantic.
A lot a bug fixes
lets go mergin

PR #146, opened 2021-09-29
Implement control and delegation permission updating

### LINKED ISSUE #145: Create bridge between client and temporal to persist control and delegation permission
We need to allow the creator of the room to update the control and delegation permission of users in the room.

### Temporal

- [x] Receive a signal with the value of the permission to set and the user id and the id of the user that wants to update the permission
- [x] Update the value `HasControlAndDelegationPermission` for the correct user
- [x] Broadcast the state to all users
- [x] Create a HTTP route

### Server

- [x] Receive a socket event with the value of the permission to set and the user id
- [x] Verify that the emitter is the creator of the room
- [x] Make an HTTP request to Temporal server with the value of the permission, the user to update and the emitter
- [x] Receive the acknowledgement from Temporal and broadcast new permissions to user's devices

### Client

- [x] Update the state machine to receive permission update events
- [x] Update the state machine `socketConnection` service to broadcast permission update events to the server

Devessier:
Closes #145.

### DIFF
--- a/packages/client/machines/appMusicPlayerMachine.ts
+++ b/packages/client/machines/appMusicPlayerMachine.ts
@@ -94,6 +94,11 @@ export type AppMusicPlayerMachineEvent =
           state: MtvWorkflowStateWithUserRelatedInformation;
       }
     | { type: 'VOTE_FOR_TRACK'; trackID: string }
+    | {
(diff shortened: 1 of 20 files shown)


prastoin (approved):
Nice deep in the backend workflow !
Lets go mergin

PR #148, opened 2021-10-01
Add server mtv room infinite loading

Devessier:
We implemented a new route to search mtv rooms by their name.
It provides pagination.


prastoin (approved):
Great work
So cool to see the project pulling out the best from lucid & adonis
Lets go mergin

PR #188, opened 2021-11-01
Begin creating a first e2e test

Devessier:


prastoin (approved):
As the previous e2e tests setup, this very first e2e test is gorgeous
Hyped to see what we're going to build with all those tools !
Ready to merge

PR #190, opened 2021-11-01
refactor(client): change minimum count to be played values

prastoin:


Devessier (approved):
Perfect!
We can merge.

PR #195, opened 2021-11-10
Create E2e Test C

Devessier:


packages/server/database/seeders/User.ts
```diff
@@ -23,19 +23,19 @@ export default class UserSeeder extends BaseSeeder {
...
             },
             {
                 uuid: '71bc3025-b765-4f84-928d-b4dca8871370',
-                nickname: internet.userName(),
+                nickname: 'Moris',
             },
             {
                 uuid: 'd125ecde-b0ee-4ab8-a488-c0e7a8dac7c5',
-                nickname: internet.userName(),
+                nickname: 'Leïla',
```
prastoin, line 34: I think setting constants names for theses users is necessary but i'm not sure about the given names
I think something more generic would be prefered
Such as available-user-A B C D etc ?

packages/client/e2e/test-c.spec.ts
```diff
@@ -0,0 +1,343 @@
+import { test, expect, Browser, Page, Locator } from '@playwright/test';
+import { assertIsNotNull, assertIsNotUndefined } from './_utils/assert';
+import { mockSearchRooms } from './_utils/mock-http';
+import { waitForYouTubeVideoToLoad } from './_utils/wait-youtube';
+
+const AVAILABLE_USERS_LIST = [
```
prastoin, line 6: Do we want to centralize this data somewhere in the new _utils folder in an other PR ?

prastoin:
Unless I'm mistaken, the following feature is not covered by the test:
```
2/ UserC goes to the mtv room search engine, on the top of the result should appears the UserA room with an invitation icon. UserC presses the mtv room card, he should see the joined room information in the mini player 
```
The test description is not explicit about it and only suggests one way to accept an invitation
I think it should be testing both toast and mtv search engine entry point
I think accepting the invitation by clicking on the invitation is the optimal way to pick as it's temporary displayed to the user.
I think that whenever we want, either by adding a new step between two existing ones or at the end, we should take a look on what's listed by default in the mtv room search engine results and normally find the invited room card

I'll be submitting an update to the basecamp e2e testing documentation file

packages/server/database/seeders/User.ts
Devessier: To me, in E2E tests we are browsing our application as a real user would do.
And a real user would not be called _available-user-A_.
Cryptography has [Alice and Bob](https://en.wikipedia.org/wiki/Alice_and_Bob#Cast_of_characters), we could have our own characters! 😆

Devessier:
Indeed, currently I only follow one path to join a room in which the user has been invited.
I think that we should not do things by halves: ensuring the user has been invited in a room is not enough.
We should fully test that a user can join a room in which she has been invited by pressing the toast, and by pressing the result in mtv search.
To do that, we should probably update the description of another test so that the creator invites a user, and the user joins the room through mtv search.

To summarize, we could press the toast in Test C, and press the room card in Mtv search results in another test, in which we would have to add an invitation step.

packages/server/database/seeders/User.ts
prastoin: Ok makes sense !
But could we be using an other than Leïla or maybe Leila as the diaeresis is quite a pain

prastoin:
I agree,
Maybe to avoid some duplication could we do that in a group test ?

prastoin (approved):
Ready to be merge after peer reviewing !
Good work

PR #204, opened 2021-11-12
Create E2e Test D

Devessier:


packages/client/e2e/test-d.spec.ts
```diff
@@ -0,0 +1,403 @@
...
+
+            setupPageFromContext(userBContext),
+        ],
+    );
+
+    const roomName = 'MusicRoom is the best';
+
+    await createPublicRoomWithInvitation({ page: userADevice1Page, roomName });
+
+    const { page: userADevice2Page } = await setupPageFromContext(userAContext);
```
prastoin, line 327: Isn't it already done line 315 to 321 ?

packages/client/e2e/test-d.spec.ts
```diff
@@ -0,0 +1,403 @@
...
+        }),
+        waitForPlayerState({
+            page: userADevice3Page,
+            testID: 'music-player-playing-device-muted',
+        }),
+        waitForPlayerState({
+            page: userBPage,
+            testID: 'music-player-playing-device-emitting',
+        }),
+    ]);
```
prastoin, line 402: Unless I'm mistaken, the userB getting the forcedDisconnection modal is not tested ?

packages/client/e2e/test-d.spec.ts
```diff
@@ -0,0 +1,403 @@
...
```
prastoin, line 385: If I'm not wrong here we should also test the emitting device eviction ?
The emitting device should be killed and one of the other device should take the lead

prastoin (approved):
Great test
Road to the last one
Merging

PR #210, opened 2021-11-16
Listen to JOINED_ROOM event globally

### LINKED ISSUE #207: Devices never join room
When a user joins a room in a device, other devices never join it.

There is an issue inside the app music player state machine that listens to this event only when it's the device that joined the room.

Devessier:
We used to wait for this event only
after having pressing a room to join it.
This will be tested in E2E tests.

Closes #207


prastoin (approved):
Great bug found while e2e testing :sunglasses: 
Ready to merge

PR #212, opened 2021-11-16
fix(temporal,client): when the only track from list is loaded as curr…

prastoin:
…ent we compare current track


prastoin:
About to create a specific test to verify this usage in temporal package

Devessier (approved):
Congratulations to you for having found the root of the bug.
We can merge!

PR #216, opened 2021-11-17
Fix localStorage check to work on mobile devices

Devessier:
The previous code was crashing on iOS simulator.
We have to use `window.localStorage`.


prastoin (approved):
:+1:

PR #218, opened 2021-11-17
Printing failing tests only using jest-silent-reporter

prastoin:


Devessier (approved):
Thanks to you we will finally be able to see which tests fail in the CI!
Great job man!

Issue #221, opened 2021-11-18
Mtv Room with constraints. Send event when time constraint starts and is done


prastoin:
If a user creates a room with constraint with startsAt in the future.
His client we'll never be directly notified via a specific event about any time constraint modification.
Which means the client tracks card won't be enabled for him until the next new state he receives from any event type.
But for a solo user in his own room he might never receives any new event to update the client about the time constraint status

This is why we need to add some timers into temporal to notify when the `timeConstraint` value is updating
Keeping in mind that @Devessier  is working on the following issue #220. That means that now startsAt can possibly be in the past

Client:
- [ ] in the appMusicPlayerMachine listen to the `'TIME_CONSTRAINT_UPDATE'` perform a assignMergeNewState in it's handler

Server:
- [ ] Create a route to handle the acknowledgeTimeConstraintUpdate temporal activity
- [ ] Persist the acknowledgeTimeConstraintUpdate to the whole room through a specific server socket event

Temporal:
- [x] During workflow creation stop checking if startsAt is in the future only, keep checking that startsAt is before endsAt
- [x] Create acknowledgeTimeConstraintUpdate that will pass a state
- [x] If startsAt is in the future create a timer from now to startsAt, it will send an acknowledgeTimeConstraintUpdate activity
- [x] create a timer from now to endsAt, it will send an acknowledgeTimeConstraintUpdate activity
- [x] Stop computing timeConstraintIsValid each time we need it, depending on start and end then created timerFuture update in the state a TimeConstraintIsValid field
- [x] Factorize the timeConstraint data validity inside a params method, and call it inside the workflow creation
- [x] tests

Types:
- [ ] Create `'TIME_CONSTRAINT_UPDATE'` server socket event with a state as payload

PR #223, opened 2021-11-18
Acknowlegde time constraint update temporal side

prastoin:
See #221


Devessier (approved):
Incredible work!
This feature will have a big impact on the human kind 😎.

PR #225, opened 2021-11-18
Acknowlegde time constraint update client

### LINKED ISSUE #221: Mtv Room with constraints. Send event when time constraint starts and is done

prastoin:
close #221

### DIFF
--- a/packages/client/__tests__/MtvRoomWithConstraints.test.tsx
+++ b/packages/client/__tests__/MtvRoomWithConstraints.test.tsx
@@ -132,7 +132,6 @@ describe('Room with constraints tests', () => {
         });
         expect(musicPlayerFullScreen).toBeTruthy();
 
-        console.log('0');
         //Track cards should be disabled
         await waitFor(() => {
             expect(screen.getByText(tracksList[0].title)).toBeTruthy();
@@ -154,7 +153,6 @@ describe('Room with constraints tests', () => {
         };
         serverSocket.emit('USER_PERMISSIONS_UPDATE', initialStateCpyA);
 
-        console.log('1');
         //Track cards should be disabled
         await waitFor(() => {
             expect(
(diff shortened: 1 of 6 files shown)


Devessier (approved):
Great job!
We can merge.

PR #230, opened 2021-11-18
refactor(temporal,server): look for last known mtv creator position w…

### LINKED ISSUE #227: Check for mtv room creator's position while creating a mtv room with constraints
When a user creates a room with constraints
The initial track won't receive his vote even if the server knows he fits the position constraints.
Atm we're computing the data we have on user's device after the room creation
We should pass in the room creation a new field `creatorFitsThePositionConstraint` which will be a bool

Client:
Nothing

Server:
- [x] pass in the createRoom body a new props called `creatorFitsThePositionConstraint`, it will be null if the room is not concerned by any constraint. False or true if it does.
- [x] `creatorFitsThePositionConstraint` will be calculated using existing `MtvRoomWsController.checkUserDevicesPositionIfRoomHasPositionConstraints` method
- [ ] Update tests

Temporal:
- [x] expect the new field in the body
- [x] Change the `initialUser.userFitsPositionConstraint` depending on the added field (no tests involved are logic is in the API)

prastoin:
…hile creating room

close #227

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,11 +41,12 @@ jobs:
                   node-version: 16.8.0
                   check-latest: true
 
+            #Modify key version to force cache reset
             - name: Cache node_modules 📦
               uses: actions/cache@v2
               with:
-                path: '**/node_modules'
-                key: ${{ runner.os }}-modules-${{ hashFiles('**/yarn.lock') }}
+                  path: '**/node_modules'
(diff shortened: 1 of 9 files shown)


Devessier (approved):
Great! It works perfectly 😎.
We can merge.

Devessier:
Again, it's a really great work!
We can merge.

PR #231, opened 2021-11-18
Client retrieve device location if permission are already granted

prastoin:
In this PR we're about to find smooth ways to improve the request device location popup in the UX


prastoin:
Might be reopen later ?

PR #236, opened 2021-11-19
Refactor temporal create api logic

### LINKED ISSUE #232: Put the create temporal api route handler logic inside the workflow itself

prastoin:
close #232


Devessier (approved):
Again, it's a really great work!
Validation is correctly tested now. :ok_hand:
We can merge.

PR #238, opened 2021-11-19
Update Device is emitting in database after join or create mtv room operation

### LINKED ISSUE #235: When device exits, emitting device is not reassigned
See https://cln.sh/R4lJR0 [French].

prastoin:
close #235


Devessier (approved):
Great bug fix!
We can merge.

PR #240, opened 2021-11-19
Always forward tracks list after suggestion

### LINKED ISSUE #239: Tracks suggested by users outside of room area are not forwarded to other users
See https://cln.sh/BYYgI3 [French].

Devessier:
Closes #239


prastoin (approved):
Great ! Totally agree with what I've read
This is a nice fix

PR #241, opened 2021-11-22
fix(server): add order_by on roomID to avoid retrieving previous page…

### LINKED ISSUE #234: Mtv room search engine issue, BB brunes static element always displayed

prastoin:
… result

close #234


Devessier (approved):
Great fix!
We don't see a `do/while` loop in use every day. 😄 
We can merge!

PR #242, opened 2021-11-22
Device coords timestamp

### LINKED ISSUE #226: Automatic eviction on device coords pg
Add an eviction to the coords lat lng column into the device table
Or save a timestamp to determine if coords are still valid or not

prastoin:
close #226


Devessier (approved):
Good job!
We can merge.

PR #243, opened 2021-11-22
Cache deep clone of tracks

Devessier:
If we assign `internalState.Tracks` to `internalState.TracksCheckForVoteUpdateLastSave`, the both sets will point to the same underlying tracks array.

We need to assign to `internalState.Tracks` a deep clone of tracks.


prastoin (approved):
Nice one !
Ready to merge

PR #244, opened 2021-11-22
E2e test f

Devessier:
This test is finally done!
The scenario has been adapted from the one described in Basecamp.

### DIFF
--- a/packages/client/e2e/_utils/page.ts
+++ b/packages/client/e2e/_utils/page.ts
@@ -92,6 +92,13 @@ export async function setupAndGetUserPage({
     };
 }
 
+export async function initPage(page: Page): Promise<void> {
+    await page.goto('/');
+
+    const focusTrap = page.locator('text="Click"').first();
+    await focusTrap.click();
+}
+
 /**
  * Use this function to create tab
  */
@@ -100,12 +107,14 @@ export async function createNewTabFromExistingContext(
 ): Promise<{ page: Page }> {
     const page = await context.newPage();
 
-    await page.goto('/');
-
-    const focusTrap = page.locator('text="Click"').first();
-    await focusTrap.click();
+    await initPage(page);
 
     return {
         page,
     };
+}
+
+export async function closeAllContexts(browser: Browser): Promise<void> {
+    const contexts = browser.contexts();
+    await Promise.all(contexts.map(async (context) => await context.close()));
 }
--- a/packages/client/e2e/_utils/page.ts
+++ b/packages/client/e2e/_utils/page.ts
@@ -79,10 +79,8 @@ export async function setupAndGetUserPage({
         context: context,
         knownSearches,
     });
-    await page.goto('/');
 
-    const focusTrap = page.locator('text="Click"').first();
-    await focusTrap.click();
+    await initPage(page);
 
     return {
         context,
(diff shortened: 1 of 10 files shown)


packages/client/e2e/_utils/page.ts
```diff
@@ -92,6 +92,13 @@ export async function setupAndGetUserPage({
     };
 }
 
+export async function initPage(page: Page): Promise<void> {
```
prastoin, line 95: I think we should also be using this function in the above function setupAndGetUserPage

packages/client/playwright.config.ts
```diff
@@ -6,6 +6,10 @@ const config: PlaywrightTestConfig = {
     use: {
         headless: false,
         baseURL: 'http://localhost:4000',
+        viewport: {
```
prastoin, line 9: Reminder: ask why

packages/client/e2e/test-f.spec.ts
```diff
@@ -0,0 +1,532 @@
...
+test('Test F', async ({ browser }) => {
+    const [
+        { page: userADevice1Page },
+        { context: userBContext, page: userBDevice1Page },
+    ] = await Promise.all([
+        setupAndGetUserPage({
+            browser,
+            userIndex: 0,
+            knownSearches,
+            town: 'Manosque, France',
```
prastoin, line 393: Jean Giono would be pride

packages/client/e2e/test-f.spec.ts
```diff
@@ -0,0 +1,532 @@
...
+
+    // User B votes for the initial track.
+    // It should have no impact.
+    // If it has one, the rest of the test will break.
+    await voteForTrackInMusicPlayerFullScreen({
+        page: userBDevice1Page,
+        trackToVoteFor: initialTrack,
+    });
+
+    // We wait for the room to allow votes according to time constraints.
```
prastoin, line 435: Unless I'm mistaken this comment should be remove

packages/client/e2e/test-f.spec.ts
```diff
@@ -0,0 +1,532 @@
...
+        page: userBDevice1Page,
+        roomName,
+    });
+
+    await voteForTrackInMusicPlayerFullScreen({
+        page: userBDevice1Page,
+        trackToVoteFor: initialTrack,
+    });
+
+    await Promise.all([
```
prastoin, line 462: Could this promise.all and the following one be one ?

packages/client/e2e/_utils/page.ts
Devessier: You're right!
I missed it.

packages/client/playwright.config.ts
Devessier: I had to modify the viewport so that the suggestion button does not hide first track cards.
Thanks to this modification, the last steps of test f can work: I'm targeting the score of the first track card, which was hidden by the suggestion button.

packages/client/e2e/test-f.spec.ts
Devessier: Yep!
It's been implemented and tested in the last commit I just pushed.

packages/client/playwright.config.ts
prastoin: Ok sounds like we should also take a decision about this button and its position in the screen

packages/client/e2e/test-f.spec.ts
prastoin: Great !

prastoin (approved):
This is a beautiful test !
Great work
Ready to merge

PR #249, opened 2021-11-24
Metro ignores server built files

### LINKED ISSUE #248: Mobile simulators can not start when server has been built
Metro bundler recognizes two `package.json` files with the same name for the server package: the original file, and its copy inside `dist` directory.
It prevents mobile simulators to launch.

Devessier:
Closes #248

See https://stackoverflow.com/a/41963217


prastoin (approved):
Looks good to me !

PR #255, opened 2021-11-25
Symbolize mtv room constraint in client

### LINKED ISSUE #187: User should be able to know if a given room has specific constraint for voting
The user should be able to know if he's about to join a room with either :
- time and position constraint
- onlyInvitedUserCanVote constraint

From a music room a user should be able to determine if he fits the room constraint he's joined
- [ ] Add an icon ??somewhere?? to notify the user that the time constraint is expired ( nothing if it's ok, as it's not depending on the user actions we want him to know only when it's false )
- [ ] Add an icon ??somewhere?? to notify the user that he need to be invited if the room is in onlyInvitedUsersCanVote.
- [ ] Add an icon ??somewhere?? to notify the user that he fits the position constraint and or doesn't fit the position constraint
- [ ] We should allow the user to know where the position constraint radius is located, in this way we should be saving coords and a radius somewhere and we might need to print a map somewhere in the front

To summary when disabling a track card in the list we should be declaring to the user why.
Then we might want to show to the user the icon in the tracks tab

prastoin:
close #187


Devessier (approved):
Amazing work!
We can merge.

PR #256, opened 2021-11-26
Fix ux issues with app search bars

### LINKED ISSUE #250: UX issues with search bars
## Issues to fix

Here are defined the issues we identified when using search bars on the application.

- [x] Pressing clear button does not submit input on `MusicTrackVoteSearchScreen`
- [x] Search bars that use `searchTrackMachine`, such as `SuggestTrackModal` screen, are deactivated after first submitting
- [x] Clearing and cancelling do not trigger fetching of UsersSearch

## `SearchTrackScreens` simplification (to be discussed)

Moved to #257.

Devessier:
Closes #250

We removed the `TouchableWithoutFeedback` component that was around screen content.

We tried to put tests for clearing and cancelling users invitation inside the test model, but we could not figure out a way that does not exponentially increases generated tests number.


prastoin (approved):
Great fix by refactoring this whole mess
It's very good to have every search engine based on the same machines
In my opiniog ready to merge !

PR #258, opened 2021-11-29
Screen superposition issue

prastoin:


Devessier (approved):
Good job @prastoin.
This bug fix was not easy.
We can merge.

prastoin:
To fix the main issue of the PR we've been using the `detachPreviousScreen: false` props on involved navigators.
It brings up a new bug described [here](https://stackoverflow.com/questions/63171131/when-rendering-iframes-with-html-android-crashes-while-navigating-back-to-s)
That we already had few months ago. The 0.99 opacity on the webView wasn't enough anymore
Had to disabled react navigation animation on the rootStack navigator

Devessier (approved):
Great!
We can merge.

PR #260, opened 2021-11-30
Leave room after room creation

### LINKED ISSUE #259: Until new room is created stay in the current one
User should be able to still listen to the current mtv room he's in and enters the mtv room creation funnel to avoid desync between a same user sessions

prastoin:
close #259 
+ e2e testing improvements


Devessier (approved):
Great work!
We can merge.

PR #261, opened 2021-11-30
Listen for room state changes even when track is loading

Devessier:
When play button was pressed too fast by another user, the player was not synced.
This is because the play event was not handled by app music player machine.


prastoin (approved):
Great fix !
Not an easy race condition to debug
Ready to merge

PR #262, opened 2021-11-30
Simplify search tracks

### LINKED ISSUE #257: Simplify SearchTracks screens
Currently there is one screen to hold an input to search for tracks, and another screen to hold results.
We implemented this feature that way because at the beginning of the project we wanted to show special information on SearchTrackScreen, such as the history of searches.
I don't think we will ever implement that, and the current implementation is not great.
We had to find a hack to reset the search bar on SearchTrackScreen after a room has been created (see #247). As a consequence, we can not go back to the search bar if the results are not satisfying because it is reset when we enter back on the screen!

Everything would be so much easier if we were using patterns we developed more recently, such as on `MusicTrackVoteSearchScreen`: a single screen that holds search bar and results. No more headache to reset the input whenever we want.

Devessier:
Closes #257 

- Mock `enableScreens` from `react-native-screens` in tests
- Call `jest.clearAllMocks()` after each test: needed to reset toast library
- `SuggestTrackResultsModal.tsx ` and `SearchTrackResultsScreen.tsx` have been removed
- Updates to `searchTrackMachine` allow to submit several times the form and to let upper machines handle what to do when clear or cancel button is pressed


prastoin (approved):
Amazing factorization work
Great decision
Ready to merge !

PR #265, opened 2021-11-30
Prevent suggest track button overlapping

### LINKED ISSUE #245: Suggest button Position might hide tracks score depending on the viewport

Devessier:
Closes #245


prastoin (approved):
Very good idea to prevent this issue!
Lets go merging

Issue #270, opened 2021-12-02
User creates MPE room


prastoin:
A MPE creation begins via a client performing a room creation, the operation is parsed handled by adonis and will be persisted or not to temporal to create new workflow etc 
This is the topic of this Issue
Here we'll described how should, adonis and temporal, handle the MPE creation operation. 
The client will be a temporary static way to perform an MPE room creation, such as a quite dumb button on the home emitting a specific socket event
Written socket event or types names may not be 100% accurate to the future code base

Server:
- [x] create a new file named `MpeWsController.ts` containing the class `MpeWsController`
- [x] create a socket handler for the client event `MPE_CREATE_ROOM` checking socket credentials as usual
- [x] Add a method named `onCreate` to the `MpeWsController` class. In this method will be parsed the given `MpeCreateRoomArgs`
If everything is okay persist the operation to temporal.
- [x] rename `ServerToTemporal.ts` by `MtvServerToTemporal.ts` idem with class inside
- [x] rename `TemporalToServer.ts` by `MtvTemporalToServer.ts` idem with class inside
- [x] create a file named `MpeServerToTemporal.ts` containing a class named as the file
- [x] create `createMpeWorkflow` method inside `MpeServerToTemporal` class
- [x] On room creation error either from parsing options nor from temporal sync error send back a `MPE_CREATE_ROOM_FAILURE`
- [x] Create a new table and lucid model for MPE rooms
- [x] Add new relationship to User ManyToMany MPE
- [x] On the Mpe model add a BelongsTo relationship to a User model
- [x] On the Mpe model add a HasMany relationship to User models on members a column
- [x] tests creation
- [x] tests creation fail

Temporal:
- [x] extract mtv routing from the `api/main.go` file into a new one `api/mtvRoutes.go` like [this](https://stackoverflow.com/questions/48612319/how-to-split-go-gorilla-mux-routes-in-2-separate-files-of-the-same-package)
- [x] Rename every previous mtv route by adding the prefix mtv e.g: `/create` becomes `mtv/create`
- [x] Create a new file `api/MpeRoutes.go`
- [x] Create and handle new `/mpe/create` route parsing args if everything is okay create a new MPE workflow
- [x] Create new files as `mpe.go` etc etc
- [x] Define a mpe workflow inside the `mpe.go` file. 
In it's internalState will be found:
- roomName
- usersLength
- initialParams (`MpeCreateRoomArgs`)
- Mpe Machine
- playlistDuration
[...]

PR #274, opened 2021-12-02
Refactor explicitly name mtv related method/events

### LINKED ISSUE #268: Rename every MTV websocket event to contain the keyword MTV
e.g `CREATE_ROOM` becomes `MTV_CREATE_ROOM`

prastoin:
close #268 
also related to #270


packages/server/app/Controllers/Http/Temporal/MtvServerToTemporalController.ts
```diff
@@ -13,7 +13,7 @@ import {
 import got from 'got';
 import urlcat from 'urlcat';
 
-const TEMPORAL_ENDPOINT = Env.get('TEMPORAL_ENDPOINT');
+const MTV_TEMPORAL_ENDPOINT = `${Env.get('TEMPORAL_ENDPOINT')}/mtv`;
```
Devessier, line 16: To be sure we encounter no issue if `TEMPORAL_ENDPOINT` environment variable ends with a slash `/` or not, we could use `urlcat` here too.

packages/server/start/routes.ts
```diff
@@ -28,71 +28,85 @@ Route.post('/search/users', 'SearchUsersController.searchUsers');
 
 Route.get('/ping', () => console.log('pong'));
 
-/// Temporal Routes ///
+Route.get('/proxy-places-api/*', 'PlacesApisController.proxyPlacesAPIRequest');
+
+/// Temporal MTV Routes ///
+
+const MTV_TEMPORAL_LISTENER = `/temporal/mtv`;
```
Devessier, line 35: I think it would be better to use AdonisJS idiomatic way to _group_ routes: https://docs.adonisjs.com/guides/routing#prefix-routes.
That way we do less handmade work.

Devessier (changes requested):
Really great job.
It's absolutely not easy to do such a massive refactoring.
I think we can replace two handmade solutions to prevent some tricky mistakes from happening.

packages/server/app/Controllers/Http/Temporal/MtvServerToTemporalController.ts
prastoin: Done

packages/server/start/routes.ts
prastoin: Done

Devessier: Perfect!

Devessier (approved):
Perfect!
We can merge.

PR #275, opened 2021-12-02
test(server): add rollback middleware everytime in japafile before ru…

prastoin:
…nning test


Devessier (approved):
Really great improvement!
We can merge.

PR #279, opened 2021-12-03
Server room creation migrations

prastoin:
related to #270


Devessier (approved):
It's a great beginning for server-side MPE!
We can merge.

PR #280, opened 2021-12-06
Improve mpe creation fail handling

prastoin:
related to #270


Devessier (approved):
Great job!
We can merge.

Issue #282, opened 2021-12-07
MPE room creation acknowledgement 


prastoin:
Server:
- [x] create a MpeTemporalToServer.ts file, containing a class MpeTemporalToserverController
- [x] Create a room creation acknowledgement handler inside MpeTemporalToserverController emit a MPE_ROOM_CREATE_CALLBACK event in the given room
- [x] tests

Temporal:
- [x] After initialTrack is fetched. Send an mpeRoomAcknowledgeMpeRoomCreation activity, containiny an exposedState
- [x] In the api createMpeWorkflow handler send back a state query result
- [x] tests

Types:
- [x] create a MPE_ROOM_CREATE_SYNCED_CALLBACK that will be triggered just after persisting the operation to temporal, the given state will be generated by temporal at the request fly, in a certain way syncly. With a payload MpeWorkflowState
- [x] create a MPE_ROOM_CREATE_CALLBACK with a payload MpeWorkflowState. Will be triggered on a temporal mpe-room-creation-acknowledgement activity response

PR #285, opened 2021-12-07
feat(temporal): mpe initialTrack is fetched send room creation acknow…

prastoin:
After initial track is fetched send Mpe room creation acknowledgement activity to the server
related to #282


Devessier (approved):
Good job.
We can merge.

Issue #283, opened 2021-12-07
MPE Tracks List Editing - Adding a track


Devessier:
## Temporal

- [x] Receive a signal with tracks IDs to add (`AddTracks`)
- [x] Keep only tracks not currently in the playlist before fetching their information
- [x] Fetch tracks information
- [x] Keep only tracks not currently in the playlist after fetching their information
- [x] If no tracks could be added (because of duplicates) send an error to the device that tried the operation (request `/mpe/reject-adding-tracks`)
- [x] If tracks could be added, forward updated tracks list to all users and acknowledge the operation (request `/mpe/acknowledge-adding-tracks`)
- [x] Create route `/mpe/add-tracks` that sends `AddTracks` signal to the workflow

## Server

- [x] Wait for `MPE_ADD_TRACKS` WS event and request `/mpe/add-tracks` route of Temporal
- [x] Create route `/mpe/reject-adding-tracks`
- [x] Send `MPE_ADD_TRACKS_FAIL_CALLBACK` WS event to the device that tried the operation
- [x] Create route `/mpe/acknowledge-adding-tracks`
- [x] Send `MPE_TRACKS_LIST_UPDATE` WS event to all devices
- [x] Send `MPE_ADD_TRACKS_SUCCESS_CALLBACK` to the device that added the track

## Client

- [x] Create a screen to pick a track to add to the playlist (`MusicPlaylistEditorSearchTrackScreen`)
- [x] When a track is picked, send a `MPE_ADD_TRACKS` event to the server
- [x] When receiving `MPE_ADD_TRACKS_FAIL_CALLBACK`, show an error toast
- [x] When receiving `MPE_ADD_TRACKS_SUCCESS_CALLBACK`, show a success toast and update tracks list
- [x] When receiving `MPE_TRACKS_LIST_UPDATE`, update tracks list

## Notes

~~We only check for duplicates after tracks to add information have been fetched because tracks can be deleted in the meantime.~~
~~_Is this really the expected behaviour?_~~

A: We remove duplicates before and after information fetching.

prastoin:
> We only check for duplicates after tracks to add information have been fetched because tracks can be deleted in the meantime.
> _Is this really the expected behaviour?_
 I think we should be doing the same as we do in the MTV.
 We check for duplicated track before fetching more information, but also after given track fetched information is finished as there is a race condition risk

prastoin:
[...]

PR #290, opened 2021-12-08
MPE Adding Tracks - Temporal

Devessier:
Temporal part of #283

### DIFF
--- a/packages/temporal/api/mpe_routes.go
+++ b/packages/temporal/api/mpe_routes.go
@@ -15,6 +15,7 @@ import (
 
 func AddMpeHandler(r *mux.Router) {
 	r.Handle("/mpe/create", http.HandlerFunc(createMpeRoomHandler)).Methods(http.MethodPut)
+	r.Handle("/mpe/add-tracks", http.HandlerFunc(MpeAddTracksHandler)).Methods(http.MethodPut)
 }
 
 type MpeCreateRoomRequestBody struct {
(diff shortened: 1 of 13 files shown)


packages/temporal/api/mpe_routes.go
```diff
@@ -100,6 +101,50 @@ func createMpeRoomHandler(w http.ResponseWriter, r *http.Request) {
...
+
+	signal := shared_mpe.NewAddTracksSignal(shared_mpe.NewAddTracksSignalArgs{
+		TracksIDs: body.TracksIDs,
+		UserID:    body.UserID,
+		DeviceID:  body.DeviceID,
+	})
+	if err := temporal.SignalWorkflow(
+		context.Background(),
+		body.WorkflowID,
+		"",
```
prastoin, line 134: should we create something like `shared.NoWorkflowRunIDRelated`

packages/temporal/mpe/workflows/mpe.go
```diff
@@ -161,6 +168,84 @@ func MpeRoomWorkflow(ctx workflow.Context, params shared_mpe.MpeRoomParameters)
...
+
+					MpeRoomAddedTracksInformationFetchedEventType: brainy.Transition{
+						Actions: brainy.Actions{
+							brainy.ActionFn(
+								func(c brainy.Context, e brainy.Event) error {
+									event := e.(MpeRoomAddedTracksInformationFetchedEvent)
+
+									// If one track to add is already in the playlist, abort the whole operation.
+									// In practice, a single track can be added at a time.
+									// Therefore, aborting the whole operation makes sense.
```
prastoin, line 222: I totally understand this point of view, this is completly ok.
I don't know if it's still the best behaviour. Even if we're not expecting any user to be able to send a tracks array, Temporal is agnostic about that. Maybe should we throw an error only if none of the fetched tracks could be added ?

packages/temporal/api/mpe_routes.go
Devessier: That's a good idea, yeah!

packages/temporal/mpe/workflows/mpe.go
Devessier: It would be better, yeah!

prastoin (approved):
Great work ! very nice
Thank for your feedback
Ready to merge

PR #295, opened 2021-12-09
MPE Adding Tracks - Server

Devessier:
Server part of #283

### DIFF
--- a/packages/server/start/mpeSocket.ts
+++ b/packages/server/start/mpeSocket.ts
@@ -56,10 +56,10 @@ export default function initMpeSocketEventListeners(socket: TypedSocket): void {
     });
 
     socket.on('MPE_ADD_TRACKS', async (rawArgs) => {
-        try {
-            const { roomID, tracksIDs } =
-                MpeRoomClientToServerAddTracksArgs.parse(rawArgs);
+        const { roomID, tracksIDs } =
+            MpeRoomClientToServerAddTracksArgs.parse(rawArgs);
 
+        try {
             const { user, deviceID } =
                 await SocketLifecycle.getSocketConnectionCredentials(socket);
 
@@ -79,7 +79,10 @@ export default function initMpeSocketEventListeners(socket: TypedSocket): void {
             });
         } catch (e) {
             console.error(e);
-            socket.emit('MPE_ADD_TRACKS_FAIL_CALLBACK');
+
+            socket.emit('MPE_ADD_TRACKS_FAIL_CALLBACK', {
+                roomID,
+            });
         }
     });
 }
--- a/packages/server/start/mpeSocket.ts
+++ b/packages/server/start/mpeSocket.ts
@@ -20,7 +20,7 @@ async function isUserInMpeRoom({
     const mpeRoom = await MpeRoom.query()
         .where('uuid', roomID)
         .andWhereHas('members', (queryUser) => {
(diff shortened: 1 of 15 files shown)


packages/types/src/mpe-temporal-to-server.ts
```diff
@@ -0,0 +1,18 @@
...
+    roomID: z.string().uuid(),
+    deviceID: z.string().uuid(),
+});
+export type MpeRejectAddingTracksRequestBody = z.infer<
+    typeof MpeRejectAddingTracksRequestBody
+>;
+
+export const MpeAcknowledgeAddingTracksRequestBody = z.object({
+    state: MpeWorkflowState,
+    deviceID: z.string().uuid(),
```
prastoin, line 14: This is not a big deal, but I think that a deviceID should always be associed to a userID
As far as for logging, keeping trace is concerned, if the device just got evicted

prastoin:
Great work from here !
Really nice usage of the japa regression tests, perfect use case
About to check with you the above raised points before approving the PR

packages/server/start/mpeSocket.ts
```diff
@@ -1,9 +1,33 @@
...
 import SocketLifecycle from 'App/Services/SocketLifecycle';
 import Ws from 'App/Services/Ws';
 import { TypedSocket } from './socket';
 
+interface IsUserInMpeRoomArgs {
+    userID: string;
+    roomID: string;
+}
+
+async function isUserInMpeRoom({
```
prastoin, line 16: Should put this logic inside the `SocketLifeCycle.getSocketConnectionCredentials` method ?
We could add an optionnal parameter to the function such as `relatedMpeRoomID`, via a custom lucid query we could return the matching mpe room or compute and return a boolean `userIsInGivenMpeRoom` ?

packages/server/tests/MpeTracksListEditing.test.ts
```diff
@@ -0,0 +1,400 @@
...
+                    AllServerToClientEvents['MPE_ADD_TRACKS_SUCCESS_CALLBACK']
+                >(noop);
+            userBSocket1.on(
+                'MPE_ADD_TRACKS_SUCCESS_CALLBACK',
+                userBSocket1AddTracksSuccessCallbackSpy,
+            );
+
+            const userBSocket1TracksListUpdateSpy =
+                sinon.spy<
+                    AllServerToClientEvents['MPE_ADD_TRACKS_SUCCESS_CALLBACK']
```
prastoin, line 375: Should this be `MPE_TRACKS_LIST_UPDATE` ?

packages/server/tests/MpeTracksListEditing.test.ts
```diff
@@ -0,0 +1,400 @@
...
+                >(noop);
+            userASocket2.on(
+                'MPE_ADD_TRACKS_SUCCESS_CALLBACK',
+                userASocket2AddTracksSuccessCallbackSpy,
+            );
+
+            const userASocket2TracksListUpdateSpy =
+                sinon.spy<
+                    AllServerToClientEvents['MPE_ADD_TRACKS_SUCCESS_CALLBACK']
```
prastoin, line 287: Should this be `MPE_TRACKS_LIST_UPDATE` ?

prastoin (approved):
We will see later concerning the userIsInMpeRoom middleware
Merging

PR #301, opened 2021-12-13
feat(temporal): add default use case to every switch statement panic …

### LINKED ISSUE #296: In golang add default case to every switch statement
Add a default case to every switch statement, inside just throw a panic for the workflow signal switch. other just raise a log error

prastoin:
…on unkown operation

close #296

### DIFF
--- a/packages/temporal/mpe/workflows/mpe.go
+++ b/packages/temporal/mpe/workflows/mpe.go
@@ -19,6 +19,7 @@ import (
 
 var (
 	ErrRoomDoesNotHaveConstraints = errors.New("room does not have constraints")
+	ErrUnknownWorflowSignal       = errors.New("encountered an unkown MPE workflow signal")
 )
 
 type MpeRoomInternalState struct {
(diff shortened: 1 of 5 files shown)


packages/temporal/mpe/workflows/mpe_test.go
```diff
@@ -244,6 +246,51 @@ func (s *CreateMpeWorkflowTestUnit) Test_CreateMpeWorkflowFetchInitialTrackFaile
 	err := s.env.GetWorkflowError()
 	s.ErrorIs(err, workflow.ErrDeadlineExceeded, "The workflow ran on an infinite loop")
 }
+
+func (s *UnitTestSuite) Test_MtvRoomPanicAfterUnkownWorkflowSignal() {
```
Devessier, line 250: Really great idea to add a test for this!

packages/temporal/mpe/workflows/mpe.go
```diff
@@ -369,6 +369,8 @@ func MpeRoomWorkflow(ctx workflow.Context, params shared_mpe.MpeRoomParameters)
 						FromIndex:        message.FromIndex,
 					}),
 				)
+			default:
+				panic("Encountered an unkown MPE workflow signal")
```
Devessier, line 373: `panic` can also take an `error` as parameter.
It allows to do not use `strings.Contains` to know what was is the cause.

Devessier (approved):
Great PR! We can feel more secure now.
We can merge.

Devessier (approved):
Great!
We can merge.

PR #302, opened 2021-12-13
test(client): e2e testing the suggest becomes vote

### LINKED ISSUE #289: MTV suggest track race condition and e2e test Suggest becomes vote
Bug description:
If no track are merge into the tracks list ***after a fetchTracksInformationAndForwardInitiatirActivity only*** the client won't be alerted in any way of the failure operation
We would need to notify the client using a suggestFailActivity

Also by discovering this bug that will only happens on a specific race condition we found that we don't really test in the e2e testing a suggest that becomes a vote:
A user suggests a song already present in the tracks list, it will make him vote for it only but will send back a response to the client to allow it to quit the loader

prastoin:
close #289

### DIFF
--- a/packages/client/e2e/test-a.spec.ts
+++ b/packages/client/e2e/test-a.spec.ts
@@ -160,7 +160,18 @@ async function joinerJoinsRoom({
     await miniPlayerWithRoomName.click();
 }
 
-async function joinerSuggestsTrack({ joinerPage }: { joinerPage: Page }) {
+async function joinerSuggestsTrack({
+    joinerPage,
+    query,
+    trackToSelect,
+}: {
+    joinerPage: Page;
+    query: string;
+    //After a long test the specific searched track to select could be
+    //Displayed a lot, picking the last nth on first matching element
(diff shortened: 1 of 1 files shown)


packages/client/e2e/test-a.spec.ts
```diff
@@ -281,7 +280,7 @@ test.afterEach(async ({ browser }) => {
     await closeAllContexts(browser);
 });
 
-test('Test A', async ({ browser }) => {
+test.only('Test A', async ({ browser }) => {
```
Devessier, line 283: We should not keep a test with `.only`.

I will suggest a PR that configures ESLint to disallow committing tests with a `.only`.

Devessier (changes requested):
It's really nice to test that behaviour.
We should just remove the use of `.only` on the test.

packages/client/e2e/test-a.spec.ts
prastoin: Great idea

Devessier (approved):
Great test.
We can merge.

PR #300, opened 2021-12-10
MPE Adding Tracks - Client

Devessier:
Last part of implementing adding tracks feature.

Closes #283

PR #305, opened 2021-12-13
Factorize suggest tracks screens

Devessier:
Must be merged after #300.

I had to modify the e2e test A as the query was returning two elements.


prastoin (approved):
Usefull refactor !
lets go mergin

PR #308, opened 2021-12-14
Server change track order mpe

### LINKED ISSUE #303: MPE change track order adonis
After implementing the temporal change track order logic we need to dev and plug the adonis one
Here we want adonis to check for user auth when performing the change track order operation and adonis to spread information on temporal activity reponse.

Adonis:
- [ ] Add a socket listener for the `MPE_CHANGE_TRACK_ORDER_UP` and `MPE_CHANGE_TRACK_ORDER_DOWN` check if user can perform an action toward the given `mpeRoomID` if yes persist the operation to temporal ( implies creation new method inside MpeServerToTemporalToController and MpeWsController.ts etc etc )
- [ ] Create route and handler for the temporal response to the change track activity
- [ ] On a failure response from temporal, send back a `MPE_CHANGE_TRACK_ORDER_FAIL_CALLBACK` event to the involved client thanks to the req's body `deviceID` and `userID`
- [ ] On a success response from temporal, send back a `MPE_CHANGE_TRACK_ORDER_SUCCESS_CALLBACK` event containing a new state to the involved client thanks to the req's body `deviceID` and `userID` but also send to every user in the `mpeRoomID` an `MPE_TRACKS_LIST_UPDATE` server socket event
- [ ] tests

Types:
- [ ] Create a client to server `MPE_CHANGE_TRACK_ORDER_DOWN` event that will take a `trackID` & `mpeRoomID` as payload
- [ ] Create a client to server `MPE_CHANGE_TRACK_ORDER_UP` event that will take a `trackID` & `mpeRoomID` as payload
- [ ] Create a server to client `MPE_CHANGE_TRACK_ORDER_SUCCESS_CALLBACK` event that will take a `trackID` & `mpeRoomID` as payload
- [ ] Create a server to client `MPE_CHANGE_TRACK_ORDER_FAIL_CALLBACK` event that will take a `trackID`  & `mpeRoomID` as payload

prastoin:
close #303

### DIFF
--- a/packages/server/start/mpeSocket.ts
+++ b/packages/server/start/mpeSocket.ts
@@ -15,7 +15,7 @@ interface IsUserInMpeRoomArgs {
     roomID: string;
 }
 
-async function throwErrorIfUserIsNotInGivenMpeRoom({
+export async function throwErrorIfUserIsNotInGivenMpeRoom({
     userID,
     roomID,
 }: IsUserInMpeRoomArgs): Promise<void> {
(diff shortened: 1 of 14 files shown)


packages/server/start/mpeSocket.ts
```diff
@@ -85,4 +85,54 @@ export default function initMpeSocketEventListeners(socket: TypedSocket): void {
...
+
+            await MpeRoomsWsController.onChangeTrackOrder(
+                {
+                    roomID,
+                    trackID,
+                    userID: user.uuid,
+                    deviceID,
+                    fromIndex,
+                    operationToApply:
+                        MpeChangeTrackOrderOperationToApply.Values.DOWN,
```
Devessier, line 105: It seems that you prefer enums over literal strings unions 👀

packages/server/tests/utils/TestUtils.ts
```diff
@@ -638,3 +638,22 @@ export function generateMpeWorkflowState(
...
+>(
+    socket: TypedTestSocket,
+    event: Event,
+): sinon.SinonSpy<
+    Parameters<AllServerToClientEvents[Event]>,
+    ReturnType<AllServerToClientEvents[Event]>
+> {
+    const customSpy = sinon.spy<AllServerToClientEvents[Event]>(noop);
+
+    socket.on(event, customSpy as any); //tmp fix ?
```
Devessier, line 657: Instead of `as any`, which should never be used as it deactivates TypeScript compiler now and for all time, we should use `@ts-expect-error <Message indicating why we need that>`.
If the error disappears in the future, the compiler will throw an error because it is excepting an error.

This is not a crucial part of our codebase, so there is no need to change it. I just wanted to point out this detail. 😄

packages/types/src/mpe-server-to-temporal.ts
```diff
@@ -22,3 +22,27 @@ export const MpeAddTracksResponseBody = z.object({
...
+    fromIndex: z.number().min(0),
+    userID: z.string(),
+    deviceID: z.string(),
+    operationToApply: MpeChangeTrackOrderOperationToApply,
+});
+export type MpeChangeTrackOrderRequestBody = z.infer<
+    typeof MpeChangeTrackOrderRequestBody
+>;
+
+export const MpeChangeTrackOrderResponseBody = z.object({
```
Devessier, line 43: I really like to have a type for each response body, even if they are all the same. If in the future the body changes, we won't have to change the type we are expecting the `got` response to be, but we will probably have to change what we do from the new body.
We know that the code using the response is likely to change, therefore it will be the first place where we will search for bugs. It's less true for the _type_ of a response body.

Devessier (changes requested):
It's perfect!
We need to update an import from `@musicroom/types`.

packages/server/start/mpeSocket.ts
```diff
@@ -13,19 +15,20 @@ interface IsUserInMpeRoomArgs {
...
     userID,
     roomID,
-}: IsUserInMpeRoomArgs): Promise<boolean> {
+}: IsUserInMpeRoomArgs): Promise<void> {
     const mpeRoom = await MpeRoom.query()
         .where('uuid', roomID)
         .andWhereHas('members', (queryUser) => {
             return queryUser.where('uuid', userID);
         })
         .first();
```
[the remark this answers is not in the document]
prastoin: Always good to update, will look forward

packages/server/start/mpeSocket.ts
prastoin: No big deal to me :p
Just been doing the same as we've done for the `MtvPlayingModes` ( it was also me lol )

Devessier (approved):
Perfect business logic, perfect tests.
We need merge that now!

PR #315, opened 2021-12-14
MPE Deleting Tracks - Temporal

### LINKED ISSUE #311: MPE Deleting a track - Temporal
Deleting tracks can not fail on Temporal side.
If the user tries to delete tracks that do not exist, this is not an issue. This behaviour can be due to tracks having been already deleted before the request is handled by Temporal.

## Todo

- [x] Receive a signal with tracks IDs to delete (`DeleteTracks`)
- [x] Ensure user has valid rights
- [x] Delete tracks from tracks list
- [x] Acknowledge the operation (request `/mpe/acknowledge-deleting-tracks`)
- [x] Create route `/mpe/delete-tracks` that sends `DeleteTracks` signal to the workflow

Devessier:
Closes #311


Devessier:
We added a test to ensure the creator can delete tracks in an open room where only invited users can vote.
We scaffolded a test to ensure users not invited can not vote in a room where only invited users can vote.

PR #317, opened 2021-12-14
test(client): add setImmediate to every server socket event to client…

prastoin:
… handler

### DIFF
--- a/packages/client/screens/MusicPlaylistEditorRoomScreen/__tests__/TracksListEditing.test.tsx
+++ b/packages/client/screens/MusicPlaylistEditorRoomScreen/__tests__/TracksListEditing.test.tsx
@@ -4,19 +4,19 @@ import Toast from 'react-native-toast-message';
 import { serverSocket } from '../../../services/websockets';
 import { db, generateTrackMetadata } from '../../../tests/data';
 import {
-    renderApp,
     fireEvent,
-    waitFor,
-    within,
-    waitForElementToBeRemoved,
     render,
+    renderApp,
+    waitFor,
+    waitForElementToBeRemoved,
+    within,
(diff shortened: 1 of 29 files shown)


prastoin:
PR should now be reviewable

packages/client/screens/MusicPlaylistEditorRoomScreen/__tests__/TracksListEditing.test.tsx
```diff
@@ -4,19 +4,19 @@ import Toast from 'react-native-toast-message';
...
-    render,
+    within,
 } from '../../../tests/tests-utils';
 
 function toTrackCardContainerTestID(id: string): string {
     return `${id}-track-card-container`;
 }
 
-function extractTrackIDFromCardContainerTestID(testID: string): string {
+export function extractTrackIDFromCardContainerTestID(testID: string): string {
```
Devessier, line 19: This is a vicious bug.
When you import something from a test file inside another test file, all the tests defined in the origin file will be duplicated in each file with the import.
If we want to use this function outside of this test file, we should put it in a file that declares no tests.

Devessier (changes requested):
Great refactoring job!
We should change where we export `extractTrackIDFromCardContainerTestID` function to don't duplicate tests in each file we import the function.

packages/client/screens/MusicPlaylistEditorRoomScreen/__tests__/TracksListEditing.test.tsx
prastoin: You're right ! fixing this

Devessier (approved):
Perfect!
We can merge.

PR #321, opened 2021-12-16
Receive initial tracks after MPE creation

### LINKED ISSUE #314: MPE Initial track not displayed on client
The initial track is displayed on client only after an other action.

We need to determine the cause.

Devessier:
Closes #314


prastoin (approved):
Nice one !

Issue #345, opened 2021-12-17
MPE retrieve room informations


prastoin:
# MPE retrieve room informations

This issue is about retrieving a given mpe room informations.
When ?: when a user hits on a mpe track card summary and wants to display the mpe room view;
The getContext will compute and add to it's response a state and a ` userIsNotInRoom` boolean.
As the user could hit the mpe summary card from an other profile mpe search engine or juste via the classic mpe search engine.

Overall from a mpe search engine, a user can possibly hit a mpeRoomSummaryCard he has already joined. In this way we should not display the JOIN button if it's already the case, this is why the `userIsNotInRoom` boolean computed on the fly

Client:
- [x] After clicking on a MpeRoomSummaryCard from either Mpe search engine, library search engine, An other user MPE search engine spawn an actor. Then proceed to display the room view skeleton until we receive the related informations
- [x] Now after spawning an actor with an empty `initialContext` emit an `MPE_GET_CONTEXT` with the related mpe room `roomID` and wait for a `MPE_GET_CONTEXT_CALLBACK` ( `FAIL` or `SUCCESS` response )
- [x] On a `MPE_GET_CONTEXT_SUCCESS_CALLBACK` merge the given state into the related actor `context.state` also merge the `userIsNotInRoom` bool in the related actor `context.userIsNotInRoom`. The view should be updated now
- [x] On a `MPE_GET_CONTEXT_FAIL_CALLBACK` we should stop the related actor and proceed to redirect the user's view to the mpe room search screen, also we should display an error toast.
- [x] Depending on the `playlist.context.userIsNotInRoom` display the `JOIN` room button or not
- [ ] #362
- [x] Display mpe room view skeleton while retrieving context
- [ ] #363

Server:
- [x] Add a listener for the client socket event `MPE_GET_CONTEXT`, we should verify if room is private that the user is invited into the room to access it's information, then persist to temporal a `getState` query to the related mpeRoom worfklow, after retrieving the state compute the `userIsNotInRoom` boolean and then emit the server socket event `MPE_GET_CONTEXT_CALLBACK`
- [x] If the getState Query fails or user doesn't have the permission we will then emit a `MPE_GET_CONTEXT_FAIL_CALLBACK` containing the related `roomID`
- [x] tests

Types:
- [x] Create a client socket event `MPE_GET_CONTEXT` that takes a `roomID`
[...]

PR #356, opened 2021-12-21
Mpe get context server

prastoin:
related to #345

### DIFF
--- a/packages/server/tests/MpeGetContext.test.ts
+++ b/packages/server/tests/MpeGetContext.test.ts
@@ -1,0 +1,421 @@
+import Database from '@ioc:Adonis/Lucid/Database';
+import MpeServerToTemporalController from 'App/Controllers/Http/Temporal/MpeServerToTemporalController';
+import { datatype, random } from 'faker';
+import test from 'japa';
+import sinon from 'sinon';
+import {
+    initTestUtils,
+    generateMpeWorkflowState,
+    createSpyOnClientSocketEvent,
+    generateArray,
+} from './utils/TestUtils';
+
+test.group('MPE get context', (group) => {
+    const {
+        createUserAndGetSocket,
+        disconnectEveryRemainingSocketConnection,
+        initSocketConnection,
+        waitFor,
+    } = initTestUtils();
+
+    group.beforeEach(async () => {
+        initSocketConnection();
+        await Database.beginGlobalTransaction();
+    });
+
+    group.afterEach(async () => {
+        await disconnectEveryRemainingSocketConnection();
+        sinon.restore();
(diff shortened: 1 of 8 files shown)


packages/server/tests/MpeGetContext.test.ts
```diff
@@ -0,0 +1,412 @@
...
+
+    group.afterEach(async () => {
+        await disconnectEveryRemainingSocketConnection();
+        sinon.restore();
+        await Database.rollbackGlobalTransaction();
+    });
+
+    test('It should retrieve context from an already joined room', async (assert) => {
+        const creatorUserID = datatype.uuid();
+        const roomsIDs = Array.from({
```
Devessier, line 33: There is a `generateArray` function in `TestUtils.ts` that we can use to create an array of random size.

packages/server/tests/MpeGetContext.test.ts
```diff
@@ -0,0 +1,412 @@
...
+        });
+        ///
+
+        //User retrieving room context
+        const roomsIDs = Array.from({
+            length: datatype.number({ min: 3, max: 7 }),
+        }).map(() => ({
+            roomID: datatype.uuid(),
+            roomName: random.words(3),
+        }));
```
Devessier, line 108: We can use `generateArray` from `TestUtils.ts`.

packages/server/tests/utils/TestUtils.ts
```diff
@@ -354,12 +355,13 @@ export function initTestUtils(): TestUtilsReturnedValue {
...
             await Promise.all(
                 mpeRoomIDToAssociate.map(
-                    async ({ roomID, roomName: mpeRoomName }) => {
+                    async ({ roomID, roomName: mpeRoomName, isOpen }) => {
                         let mpeRoomToAssociate = await MpeRoom.find(roomID);
                         if (mpeRoomToAssociate === null) {
                             console.log('CREATING MPE ROOM FROM TEST UTILS');
                             mpeRoomToAssociate = await MpeRoom.create({
                                 uuid: roomID,
+                                isOpen: isOpen !== undefined ? isOpen : true,
```
Devessier, line 364: This is just a coding style concern and does not need to be changed, but this can also be written like this, using [nullish coalescing operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator):

```suggestion
                                isOpen: isOpen ?? true,
```

Devessier (changes requested):
Very good job!

I think you should change how you create random arrays in your tests so that we keep consistent patterns in our tests.

Devessier (approved):
Great job!
We can merge.

PR #357, opened 2021-12-21
Mpe get context client

prastoin:
realted to #345


prastoin:
From there I was wondering how should I be testing the code I've made without going neither too much or not enough deep.
This dev would be fully coverable via the Library mpe search engine feature, via mpeSummary etc etc
In my opinion I should build a simple communication with the server to retrieve every user's mpe room, that could later be the foundation of the final mpe search engine server ?

prastoin:
Also I still need to implem the spawn already joined room actor on related server socket event reception feature

packages/client/machines/appMusicPlaylistsMachine.ts
```diff
@@ -94,11 +107,36 @@ export const appMusicPlaylistsModel = createModel(
...
+            if (actor === undefined) {
+                console.error(
+                    'encountered unkown roomID in stop playlist actor',
+                );
+                return playlistsActorsRefs;
+            }
+
+            if (actor.ref.stop) {
+                actor.ref.stop();
+            }
```
Devessier, line 126: To stop an actor we should be using [`stop`](https://xstate.js.org/docs/guides/actors.html#stopping-actors) action.
As a consequence, we would have a first action to stop the actor, and another one to remove it from the context.

Actually, the less we interact with the actor imperatively inside the machine, the best it is.

packages/client/machines/appMusicPlaylistsMachine.ts
```diff
@@ -114,6 +152,28 @@ const spawnPlaylistActor = appMusicPlaylistsModel.assign(
...
+                    name: getPlaylistMachineActorName(roomID),
+                }),
+            };
+
+            return [...playlistsActorsRefs, playlist];
+        },
+    },
+    'SPAWN_NEW_PLAYLIST_ACTOR_FROM_ROOM_ID',
+);
+
```
Devessier, line 176: 👍

packages/client/machines/playlistMachine.ts
```diff
@@ -160,23 +163,59 @@ export type PlaylistMachineEvents = EventFrom<typeof playlistModel>;
...
+                            },
+
+                            target: 'retrievingContext',
+                        },
+                        {
+                            target: 'idle',
+                        },
+                    ],
+                },
+            },
```
Devessier, line 202: It seems that what you wanted is an [_eventless transition_](https://xstate.js.org/docs/guides/transitions.html#eventless-always-transitions).

packages/client/machines/playlistMachine.ts
```diff
@@ -160,23 +163,59 @@ export type PlaylistMachineEvents = EventFrom<typeof playlistModel>;
...
+                    });
+                }),
+
+                on: {
+                    ASSIGN_MERGE_NEW_STATE: {
+                        actions: assignMergeNewState,
+                        target: 'idle',
+                    },
+                },
+            },
```
Devessier, line 217: Amazing asynchronous communication with parent machine!

packages/client/machines/playlistMachine.ts
Devessier: Another pattern would to be dynamically compute the _initial_ state of the machine, depending on the parameter the machine creator received.

Devessier (changes requested):
Good job!
I wrote some thoughts about XState patterns.

packages/server/tests/MpeSearchEngine.test.ts
```diff
@@ -0,0 +1,148 @@
...
+        await Database.beginGlobalTransaction();
+    });
+
+    group.afterEach(async () => {
+        await disconnectEveryRemainingSocketConnection();
+        sinon.restore();
+        await Database.rollbackGlobalTransaction();
+    });
+
+    test('Sends updated tracks list to all other users if deleting tracks succeeded', async (assert) => {
```
Devessier, line 30: Is the test name accurate?

packages/server/tests/MpeSearchEngine.test.ts
```diff
@@ -0,0 +1,148 @@
...
+import test from 'japa';
+import sinon from 'sinon';
+import supertest from 'supertest';
+import {
+    LibraryMpeRoomSearchResponseBody,
+    MpeRoomSearchRequestBody,
+} from '@musicroom/types';
+import { BASE_URL, initTestUtils, generateArray } from './utils/TestUtils';
+
+test.group('MPE Delete Tracks', (group) => {
```
Devessier, line 12: Is the tests group name accurate?

packages/client/screens/MusicPlaylistEditorRoomsSearchScreen/index.tsx
```diff
@@ -0,0 +1,118 @@
...
+                roomID,
+                roomName,
+            });
+        }
+
+        useEffect(() => {
+            libraryRoomsSend({
+                type: 'FETCH_ROOMS',
+            });
+        }, [libraryRoomsSend]);
```
Devessier, line 71: The purpose of this `useEffect` is to fetch rooms each time we go back to the screen?

packages/client/machines/appMusicPlaylistsMachine.ts
```diff
@@ -400,6 +482,47 @@ export function createAppMusicPlaylistsMachine({
...
+                                (_, { roomID }) =>
+                                    appMusicPlaylistsModel.events.REMOVE_ACTOR_FROM_PLAYLIST_LIST(
+                                        {
+                                            roomID,
+                                        },
+                                    ),
+                                {
+                                    delay: 100,
+                                },
+                            ),
```
Devessier, line 507: I think we should not rely on an arbitrary timeout.
Replacing error throwing in `usePlaylist` hook with a return type of either the actor or undefined could be a solution.

packages/client/jest.setup.tsx
```diff
@@ -323,5 +323,5 @@ beforeEach(() => {
     });
 });
 
-// jest.spyOn(console, 'warn').mockImplementation();
-// jest.spyOn(console, 'error').mockImplementation();
+jest.spyOn(console, 'warn').mockImplementation();
+jest.spyOn(console, 'error').mockImplementation();
```
Devessier, line 327: Do we keep these lines uncommented?

Devessier (dismissed):
Very heavy and great job!
Let's talk about what I commented before merging.

packages/client/screens/MusicPlaylistEditorRoomsSearchScreen/index.tsx
prastoin: Yes

packages/client/machines/appMusicPlaylistsMachine.ts
prastoin: Yes will look forward

packages/client/screens/MusicPlaylistEditorRoomsSearchScreen/index.tsx
Devessier: Okay!
I think the `useEffect` is not necessary because the initial state of the machine is `fetchingRooms` and the state machine will be interpreted by `useMachine` each time the screen is created, and will be stopped when the component is destroyed. Those are the lines in `useInterpret`, used inside `useMachine`, that deal with starting and stopping the service: <https://github.com/statelyai/xstate/blob/89fcc6d47f1890f09752d73d086fdd158898c194/packages/xstate-react/src/useInterpret.ts#L101-L120>.

That's not a big deal, I just wanted to understand this part of the screen code.

prastoin: I'm not sure this will be the case here, as we're exporting const machines instances
I've done this as I've been struggling on this point, if I only export a createMachine functionn between each rerender the value passed to useMachine will be different and xState will raise warnings

```ts
export const libraryMpeRoomSearchMachine = createMpeRoomUniversalSearchMachine({
    fetchMpeRooms: fetchLibraryMpeRooms,
});

export const mpeRoomSearchMachine = createMpeRoomUniversalSearchMachine({
    fetchMpeRooms: fetchAllMpeRooms,
});

```

Devessier: Indeed, state machines _definitions_ need to be consistent between renders.
But state machines we provide to `useMachine` or `useIntepret` hooks are only definitions. They do nothing by themselves, they need to be _interpreted_.
This is what `useMachine` and `useInterpret` do, they interpret the machine definition into a _service_ and keep it alive for all the life of the component they are called in.

Here is a demonstration without the `useEffect`: <https://cln.sh/sNh8n7>

packages/client/screens/MusicPlaylistEditorRoomScreen/index.tsx
```diff
@@ -301,6 +346,35 @@ const MusicPlaylistEditorRoomWrapper: React.FC<MpeTabMpeRoomScreenProps> = (
...
+                        style={{ flex: 1 }}
+                    >
+                        <Skeleton show={true} colorMode="dark" width="100%" />
+
+                        <Skeleton show={true} colorMode="dark" width="100%" />
+                    </MotiView>
+                </AppScreenContainer>
+            </AppScreen>
+        );
+    }
```
Devessier, line 377: Really great!

packages/client/screens/MusicPlaylistEditorRoomsSearchScreen/index.tsx
prastoin: Indeed you're right !
I've been struggling on displaying the mpe-room-card results in tests, I assume I've fixed the problem in // and kept in mind this was this specific usage I've been setting up that was solving the issue
Thanks, about to update on that point

Devessier (approved):
So great!
We can merge.

PR #360, opened 2021-12-28
Join mpe temporal

### LINKED ISSUE #336: Join MPE room temporal
# Temporal implementation of the JOIN MPE ROOM feature

- [x] Create and handle a new route `/mpe/join` that will take a `workflowID` we wanna join a `userID` `hasBeenInvited`
- [x] Create and handle a new mpe_workflow signal `join-mpe-room-signal` that takes as payload an `mpeInternalStateUser`
- [x] Create a new brainy event and listeners inside the `roomIsReady` state that will handle the join signal, inside this handler we will add the given `internalStateUser` to the `internalState.users` map, we will have to double check if the user is not already in the room, there's no reason to throw any error on that.
- [x] Normally the join workflow operation cannot fail into temporal context, in this way we will only send back a `MpeAcknowledgeJoinRoomActivity` after each `joinRoomOperation` that will take a fresh state.
- [x] tests

prastoin:
close #336


Devessier (approved):
Great job! It will be the first PR we merge in 2022.

PR #365, opened 2022-01-03
Leave mpe room temporal

### LINKED ISSUE #340: Leave MPE room temporal
# Temporal implementation of the LEAVE MPE ROOM feature

- [x] Create and handle a new route `/mpe/leave` that will take a `workflowID` we wanna leave and a `userID`
- [x] Create and handle a new mpe_workflow signal `leave-mpe-room-signal` that takes as payload an `userID`
- [x] Create a new brainy event and listeners inside the `roomIsReady` state that will handle the `leave signal`, inside this handler we will remove the given `internalStateUser` of the `internalState.users` map, we will have to double check if the user correctly in the room before, there's no reason to throw any error on that.
- [x] Normally the join workflow operation cannot fail in the temporal context, in this way we will only send back a `MpeAcknowledgeLeaveRoomActivity` after each `leaveRoomOperation` that will take a fresh state if the deletion is succesful.
- [x] Create and handle a new route `/mpe/terminate` that will take a `worfklowID` we wanna terminate
- [x] Create a new `terminateSignal`, in it's handler set `true` the `terminateWorkflow` boolean that will lead to leave the infinite loop
- [x] Tests

prastoin:
close #340


Devessier (approved):
Great job! The more I read Temporal code, the more I love this technology.
We can merge.

PR #366, opened 2022-01-03
Leave mpe room server

### LINKED ISSUE #339: Leave MPE room server
# Server implementation of the LEAVE MPE ROOM feature

Server:
- [x] Add a listener & handler for the client socket event `MPE_LEAVE_ROOM`, if the user is in the given room persist the operation to temporal, else just silent error the operation
 - If the leaving user is a basic user, persist the operation to temporal the using the `leaveWorkflow`
 - If the leaving user is the creator, persist a `terminate` workflow operation to temporal
- [x] Create a new temporal to server mpe routes `/mpe/acknowledge-leave-room` where we will spread a `MPE_USERS_LENGTH_UPDATE`  to every room's members
Do not forget to dissociate the related user model to the related MPE room model
- [x] In the `MpeWsController.onLeave` method, do the same as we do in the `MtvWsController.onLeave`:
 - Check if user is creator
 - If yes persist the call `MpeWsController.onTerminate` method that will lead to immediately disconnect every socket from the related socket io room and dissociate every room's members Model from the removed MpeRoom Model and finally persist the operation to temporal and emit in the room an `MPE_FORCED_DISCONNECTION` with the terminated `roomID`
 - If no just use `MpeServerToTemporal.leaveWorkflow` that will lead to remove every user's devices from the related socket io room and dissociate `leavingUser` model from `leavedMpeRoom`  and emit to them a `MPE_LEAVE_ROOM_CALLBACK` server socket event then persist the operation to temporal

Types:
- [x] Create `MPE_LEAVE_ROOM` client socket event that will take a `roomID`
- [x] Create `MPE_LEAVE_ROOM_CALLBACK` server socket event that takes a `roomID`
- [x] Create `MPE_FORCED_DISCONNECTION` server socket event that takes a `roomID`

prastoin:
close #339

### DIFF
--- a/packages/server/database/migrations/1638530068635_users_mpe_rooms.ts
+++ b/packages/server/database/migrations/1638530068635_users_mpe_rooms.ts
@@ -6,8 +6,14 @@ export default class UsersMpeRooms extends BaseSchema {
     public up(): void {
         this.schema.createTable(this.tableName, (table) => {
             table.increments('id');
-            table.uuid('user_uuid').references('users.uuid');
-            table.uuid('mpe_room_uuid').references('mpe_rooms.uuid');
+            table
+                .uuid('user_uuid')
+                .references('users.uuid')
+                .onDelete('CASCADE');
+            table
+                .uuid('mpe_room_uuid')
+                .references('mpe_rooms.uuid')
+                .onDelete('CASCADE');
             table.unique(['user_uuid', 'mpe_room_uuid']);
             /**
              * Uses timestamptz for PostgreSQL and DATETIME2 for MSSQL
--- a/packages/server/tests/MpeLeaveRoom.test.ts
+++ b/packages/server/tests/MpeLeaveRoom.test.ts
@@ -1,0 +1,211 @@
+import Database from '@ioc:Adonis/Lucid/Database';
+import { MpeAcknowledgeLeaveRequestBody } from '@musicroom/types';
+import MpeServerToTemporalController from 'App/Controllers/Http/Temporal/MpeServerToTemporalController';
+import { datatype, random } from 'faker';
+import test from 'japa';
+import sinon from 'sinon';
+import supertest from 'supertest';
+import urlcat from 'urlcat';
+import SocketLifecycle from 'App/Services/SocketLifecycle';
+import { MPE_TEMPORAL_LISTENER } from '../start/routes';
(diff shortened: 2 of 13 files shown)


packages/server/database/migrations/1638530068635_users_mpe_rooms.ts
```diff
@@ -6,8 +6,14 @@ export default class UsersMpeRooms extends BaseSchema {
...
-            table.uuid('user_uuid').references('users.uuid');
-            table.uuid('mpe_room_uuid').references('mpe_rooms.uuid');
+            table
+                .uuid('user_uuid')
+                .references('users.uuid')
+                .onDelete('CASCADE');
+            table
+                .uuid('mpe_room_uuid')
+                .references('mpe_rooms.uuid')
+                .onDelete('CASCADE');
```
Devessier, line 16: This won't cause issues as we have not deployed our backend to production, but a migration should never be modified.

Devessier: To do that properly, we should create a new migration, that deletes the foreign key constraint and recreate it with the `ON DELETE` constraint you want. As shown here, without a ORM: https://stackoverflow.com/a/10356720.

packages/server/tests/MpeLeaveRoom.test.ts
```diff
@@ -0,0 +1,360 @@
...
+                            .post(
+                                urlcat(
+                                    MPE_TEMPORAL_LISTENER,
+                                    'acknowledge-leave',
+                                ),
+                            )
+                            .send(body)
+                            .expect(200);
+                    });
+                });
```
Devessier, line 88: What is the purpose of these nested `setImmediate`?

packages/types/src/mpe-room-websockets.ts
```diff
@@ -1,6 +1,8 @@
 import * as z from 'zod';
 import { MpeWorkflowState } from './mpe';
+import { MpeRoomSummary } from '.';
```
Devessier, line 3: Does this import from `.` cause circular dependencies?

Devessier (dismissed):
Great job!
I wrote some comments to better understand some parts of your work.

Devessier (approved):
Great, we can merge!

PR #372, opened 2022-01-06
Leave mpe room and join e2e tests

### LINKED ISSUE #338: Leave MPE room client
# Client implementation of the LEAVE MPE ROOM feature

We should keep in mind that the leave button will be accessible from a bottom sheet modal via a settings button in the mpe room view.
From now we just want a static button in the mpe room view that will lead to perform a `MPE_LEAVE_ROOM` of the currently viewing mpe room

- [x] Hack a leave mpe room button in the related mpe room view
- [x] Add a server socket event listerner into the `AppMusicPlaylistsMachine` `socketConnection` service for the `MPE_LEAVE_ROOM_CALLBACK` that will lead to stop the related mpe Actor from the event payload `roomID` and will redirect the user view to it's library tab
- [x] Create a new Event pair between an `playlistMachine` and it's parent `AppMusicPlaylistsMachine` for the `MPE_JOIN_ROOM` workflow. Event will be forwarded to the parent then to the `SocketConnection` service achieving to sending a `MPE_JOIN_ROOM` client socket event to the server
- [x] Add a server socket event listerner into the `AppMusicPlaylistsMachine` `socketConnection` service for the `MPE_FORCED_DISCONNECTION` that will lead to stop the related mpe Actor from the event payload `roomID` and redirect to home
- [x] After a forced disconnection or a leave callback only client currently displaying the related mpe room view should be redirected
- [x] tests
- [x] e2e test b and c

prastoin:
close #338


packages/client/e2e/_utils/mpe-e2e-utils.ts
```diff
@@ -0,0 +1,642 @@
...
+    //Other users
+    forcedDisconnectedRedirectedOtherUsersPages =
+        forcedDisconnectedRedirectedOtherUsersPages
+            ? forcedDisconnectedRedirectedOtherUsersPages
+            : [];
+
+    forcedDisconnectedNotRedirectedOtherUsersPages =
+        forcedDisconnectedNotRedirectedOtherUsersPages
+            ? forcedDisconnectedNotRedirectedOtherUsersPages
+            : [];
```
Devessier, line 355:
```suggestion
export async function leaveMpeRoom({
    leavingPage,
    roomName,
    forcedDisconnectedRedirectedOtherUsersPages = [],
    forcedDisconnectedNotRedirectedOtherUsersPages = [],
    redirectedLeavingUserPages = [],
    notRedirectedLeavingUserPages = [],
}: {
    leavingPage: Page;
    roomName: string;
    forcedDisconnectedRedirectedOtherUsersPages?: Page[];
    forcedDisconnectedNotRedirectedOtherUsersPages?: NotRedirectedPage[];
    redirectedLeavingUserPages?: Page[];
    notRedirectedLeavingUserPages?: NotRedirectedPage[];
}): Promise<void> {
    // In our testing leave method, leaving page is always redirected
    const redirectedLeavingUserPagesWithLeavingPage = [leavingPage, ...redirectedLeavingUserPages];
```

Devessier: In my suggestion, we should be using `redirectedLeavingUserPagesWithLeavingPage` and not any more `redirectedLeavingUserPages` after we used it to compute `redirectedLeavingUserPagesWithLeavingPage`.

Devessier (approved):
It works perfectly!
I suggested another way to write a function.
We can merge.

packages/client/e2e/_utils/mpe-e2e-utils.ts
prastoin:
```ts
    forcedDisconnectedRedirectedOtherUsersPages = [],
    forcedDisconnectedNotRedirectedOtherUsersPages = [],
    redirectedLeavingUserPages = [],
    notRedirectedLeavingUserPages = [],
```
You're so damn right, I was wondering what native option I've missed, there we are. Thanks

Checking deeper about the `redirectedLeavingUserPagesWithLeavingPage`

Devessier (approved):
Great!
We can merge.

Issue #352, opened 2021-12-20
User receives MPE room invitation server


prastoin:
# Server implem of the User receives MPE room invitation

Server:
- [x] Refactor the `MpeWsController.onJoin` method to check for related room & user  `mpeRoomInvitation`
- [x] Update the `MpeServerToTemporal.joinWorkflow` method to take an `userHasBeenInvited` boolean
- [x] update the `MpeWsController.onCreatorInvitesUser` to emit to the invited user the `MPE_RECEIVED_ROOM_INVITATION`
- [x] tests

Below will be done during mpe search engine implem
- [ ] Update the Mpe room search engine to retrieve and sort rooms to render the following list:
- Private & invited
- Public & invited
- Public

Types:
- [x] `MPE_RECEIVED_ROOM_INVITATION` that takes `MpeRoomSummary`
- [x] Create `MpeRoomSummary`, a `MpeRoomSummary` is always user related

```ts
export const MpeRoomSummary = z.object({
    roomID: z.string(),
    roomName: z.string(),
    creatorName: z.string(),
    isInvited: z.boolean(), //Requesting user related
    isOpen: z.boolean(),
});
```

PR #373, opened 2022-01-07
Mpe invitations creator server

### LINKED ISSUE #348: Creator invites user in MPE room server
# Server implem of creator invites user in it's mpe room

Server:
- [x] Create a new table called `mpe_room_invitations` that will have containing a `invitedUserID` `invitingUserID` `mpeRoomID` all referencing an other db entry
- [x] Create a MpeRoomInvitations lucid model that will have a 3 BelongsTo relationship toward invitedUser invitingUser User Models and the MpeRoom model
- [x] Add a socket listener and handler to the `MPE_CREATOR_INVITE_USER`, in handler verify `invitingUser` is the creator of the given `mpeRoomID` and that the given user exists. Then check if an corresponding invitation is already in base.
All described errors below will be silent errors
If everything is ok then create the new MpeRoomInvitation
- [x] refactor the mpe get context to verify when encountering a private room that the requesting user is invited in the given private room
- [x] tests

Types:
- [x] `MPE_CREATOR_INVITE_USER` client socket event that takes a `invitedUserID` and a `roomID`

prastoin:
close #348 
also related to #352

### DIFF
--- a/packages/server/app/Controllers/Ws/MpeRoomsWsController.ts
+++ b/packages/server/app/Controllers/Ws/MpeRoomsWsController.ts
@@ -22,6 +22,22 @@ interface IsUserInMpeRoomArgs {
     roomID: string;
 }
 
+export async function isUserInvitedInMpeRoom({
+    roomID,
+    userID,
+    creatorID,
+}: {
+    roomID: string;
+    userID: string;
+    creatorID: string;
+}): Promise<boolean> {
+    const relatedInvitationArray = await MpeRoomInvitation.query()
+        .where('invited_user_id', userID)
+        .andWhere('mpe_room_id', roomID)
+        .andWhere('inviting_user_id', creatorID);
+    return relatedInvitationArray.length > 0;
+}
+
 export async function fromMpeRoomToMpeRoomSummary({
     room,
     userID,
@@ -35,11 +51,11 @@ export async function fromMpeRoomToMpeRoomSummary({
         'should never occurs room.creator is null',
     );
 
-    const relatedInvitationArray = await MpeRoomInvitation.query()
-        .where('invited_user_id', userID)
-        .andWhere('mpe_room_id', room.uuid)
-        .andWhere('inviting_user_id', room.creator.uuid);
-    const isInvited = relatedInvitationArray.length > 0;
+    const isInvited = await isUserInvitedInMpeRoom({
+        creatorID: room.creatorID,
+        roomID: room.uuid,
+        userID,
+    });
 
     const { isOpen, uuid: roomID, name: roomName } = room;
     const roomSummary: MpeRoomSummary = {
@@ -260,7 +276,7 @@ export default class MpeRoomsWsController {
             throw new Error('Join mpe room user is already in room');
         }
 
-        //TODO MpeRoomInvitations verifications as for as MTV
+        //TODO MpeRomInvitations verifications as for as MTV during user receives invitation feature implem
 
         const userHasBeenInvited = false;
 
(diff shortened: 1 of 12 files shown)


packages/server/database/migrations/1641557073164_mpe_room_invitations.ts
```diff
@@ -0,0 +1,36 @@
...
+                .notNullable()
+                .references('users.uuid')
+                .onDelete('CASCADE'); //always roomID creator
+            table.timestamps(true, true);
+            table.unique([
+                'mpe_room_id',
+                'invited_user_id',
+                'inviting_user_id',
+            ]);
+        });
```
Devessier, line 30: It's a pleasure to read!

Devessier (commented):
It seems perfect to me.
I will review it another time on Monday.

packages/server/app/Controllers/Ws/MpeRoomsWsController.ts
```diff
@@ -407,4 +488,68 @@ export default class MpeRoomsWsController {
...
+        await UserService.emitEventInEveryDeviceUser(
+            invitedUserID,
+            'MPE_RECEIVED_ROOM_INVITATION',
+            [
+                {
+                    roomSummary,
+                },
+            ],
+        );
+    }
```
Devessier, line 554: If I understand correctly this function, we allow the creator to invite the same user several times?
I see that we were doing the same for MTV rooms.

Devessier: Yes, it has been confirmed by the test _It should create only one invitation in base and send MPE_RECEIVED_ROOM_INVITATION to the invited user clients_ you wrote 👌

packages/server/tests/MpeRoomCreateInvitation.test.ts
```diff
@@ -0,0 +1,288 @@
...
+                invitedUserID: invitedUserID,
+                roomID,
+            });
+            assert.isTrue(false);
+        } catch ({ message }) {
+            assert.equal((await MpeRoomInvitation.all()).length, 0);
+            assert.equal(
+                message,
+                'Emitter user does not appear to be the mpe room creator',
+            );
```
Devessier, line 190: Japa has a `instanceOf` assertion that we could use to check the thrown error is an instance of a custom `Error` class we would have written.
As we did in Go, it would allow assertions to do not break if the message of the error changes, because this is not necessarily what we want to test there.
Creating custom errors is quite boring and I am not saying we should refactor this code. I just wanted to point it out.

Devessier (approved):
It's perfect! We can merge.
I wrote some comments that are just notices.

packages/server/app/Controllers/Ws/MpeRoomsWsController.ts
prastoin: Yes has we don't retrieve in the client the already invited users for a given room we still want to allow a creator to send several toast to the invited user, in reality this could bring up a toast spam issue from a malicious user creator that would scan the db
This could be changed easily by avoiding sending a toast is the invitation is already existing
But should we ?

packages/server/tests/MpeRoomCreateInvitation.test.ts
prastoin: You're completely right

packages/server/app/Controllers/Ws/MpeRoomsWsController.ts
Devessier: At the moment I think it's perfect as is, but I was also thinking about this spamming issue.
Let's go on for now!

Issue #343, opened 2021-12-17
MPE Library Search Engine - Server


Devessier:
## Notes

We do not take care of invitations in this search engine.

We must adapt code @prastoin already implemented in `MpeRoomsHttpController.listAllUserRooms`.

## Todo

- [x] This method, if receiving a `searchQuery`, filters rooms by it
- [x] Paginate results

Issue #342, opened 2021-12-17
MPE Library Search Engine - Client


Devessier:
## Todo

- [x] Forward search query to backend
- [x] Handle pagination
- [x] Model integration tests

prastoin:
>  Spawn a Playlist actor when user presses MPE room card, then redirect to the room screen

Or also if the `AppMusicPlaylistsMachine` receives an already joined and not already spawned actor MPE event

Issue #341, opened 2021-12-17
MPE Library Search Engine - Plan


Devessier:
## Behaviours

Users can scroll down to see all the rooms they joined.
There is infinite loading. There is a search bar to filter results, that uses infinite loading too.

When users press MPE rooms cards, they are redirected to the screen of the room.

## Implementation

See https://github.com/AdonisEnProvence/MusicRoom/issues/331 about Playlist actors spawning.

## Todo

- [x] #342
- [x] #343
- [x] https://github.com/AdonisEnProvence/MusicRoom/issues/387
- [x] https://github.com/AdonisEnProvence/MusicRoom/issues/388

prastoin:
All tasks have been done, closing plan issue

PR #389, opened 2022-01-14
Implement filtering and infinite loading to MPE Search Engine

Devessier:
Closes https://github.com/AdonisEnProvence/MusicRoom/issues/343
Closes https://github.com/AdonisEnProvence/MusicRoom/issues/342

Part of https://github.com/AdonisEnProvence/MusicRoom/issues/341

### DIFF
--- a/packages/client/services/MpeService.ts
+++ b/packages/client/services/MpeService.ts
@@ -2,6 +2,7 @@ import {
     MpeSearchMyRoomsRequestBody,
     MpeSearchMyRoomsResponseBody,
     ListAllMpeRoomsResponseBody,
+    ListAllMpeRoomsRequestBody,
 } from '@musicroom/types';
 import urlcat from 'urlcat';
 import redaxios from 'redaxios';
@@ -20,6 +21,7 @@ export async function fetchLibraryMpeRooms({
     const body: MpeSearchMyRoomsRequestBody = {
         userID,
         searchQuery,
+        page: 1,
     };
 
     const rawResponse = await redaxios.post(url, body);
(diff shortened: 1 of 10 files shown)


packages/server/app/Controllers/Http/MpeRoomsHttpController.ts
```diff
@@ -1,45 +1,105 @@
...
+        const mpeRoomsPagination = await user
+            .related('mpeRooms')
+            .query()
+            .where('name', 'ilike', `${searchQuery}%`)
+            .orderBy([
+                {
+                    column: 'mpe_rooms.is_open',
+                    order: 'asc',
+                },
+                // FIXME: need to handle invitations
```
prastoin, line 78: Do we want the personal user rooms to be ordered by potential invitation ?

packages/client/tests/data/index.ts
```diff
@@ -30,7 +30,7 @@ export const db = factory({
 
     searchableMpeRooms: {
         roomID: primaryKey(() => datatype.uuid()),
-        roomName: () => random.words(datatype.number({ min: 1, max: 5 })),
+        roomName: () => random.words(),
```
prastoin, line 33: :thumbsup:

prastoin: should we be doing the same for the same file line 25 ?

packages/client/services/MpeService.ts
```diff
@@ -1,41 +1,54 @@
...
-    });
-    const parsedResponse = LibraryMpeRoomSearchResponseBody.parse(
-        rawResponse.data,
-    );
+        searchQuery,
+        page,
+    };
+
+    const rawResponse = await redaxios.post(url, body);
+    const parsedResponse = MpeSearchMyRoomsResponseBody.parse(rawResponse.data);
```
prastoin, line 30: This is a real pleasure to red
So few lines for a lot of work

prastoin (approved):
Looks great !
Hail to the state machines

packages/server/app/Controllers/Http/MpeRoomsHttpController.ts
Devessier: I checked our Basecamp documentation, and it seems that we don't, indeed.
Do we even want to filter by opening status?

packages/client/tests/data/index.ts
Devessier: Why not! In another PR?

packages/client/services/MpeService.ts
Devessier: Haha yeah! Love it too.

packages/server/app/Controllers/Http/MpeRoomsHttpController.ts
prastoin: Mhm I don't think so

Issue #375, opened 2022-01-10
Fullscreen player is dismissed when loading another MTV room in the player


Devessier:
## The Workflow

A user has a MTV room in its player.

This user exports a MPE room into a MTV room.

We would expect the music player to be open full-screen with the newly created MTV room. Instead the player is not open full-screen.

## Notes

At time of writing, when a MPE room is exported into a MTV we open the player in full-screen

`leaveRoomFromLeaveRoomButton` action is called, which dismisses the full-screen player and redirects the user to the home screen.

prastoin:
Mhm I guess that displaying the musicPlayer isn't trigerrig the following cleanup as it's a hacked modal
```ts
        useFocusEffect(
            React.useCallback(() => {
                // Do something when the screen is focused
                playlistRef.send({
                    type: 'MPE_ROOM_VIEW_FOCUS',
                });

                return () => {
                    // Do something when the screen is unfocused
                    // Useful for cleanup functions
                    playlistRef.send({
                        type: 'MPE_ROOM_VIEW_BLUR',
                    });
                };
            }, [playlistRef]),
        );
```

prastoin:
Working on it

prastoin:
We should determined what behavior we would expect from the app
Do we want to expand the MusicPlayer and do not move the user preivous page ?

prastoin:
This issue led to more general issue concerning the music player fullscreen expansion after creating mtv room.
If we add the following this might also solve this issue:
-After creating a mtv room, precisely after receiving a MTV_ROOM_SYNCED_CALLBACK or a MTV_ROOM_CREATE_CALLBACK, we should toggle on the music player fullscreen

prastoin:
We determine that this issue needed to be solved
The client should be self aware that he's about to received a create room server ack from an mpe export operation, that will lead to display the created mtv room fullscreen

We still need to find a way to do so in the client

PR #390, opened 2022-01-14
avoid redirecting on home after leaving mtv room

prastoin:
close https://github.com/AdonisEnProvence/MusicRoom/issues/375


prastoin:
needs updated and about to be done there https://github.com/AdonisEnProvence/MusicRoom/pull/393

Issue #386, opened 2022-01-14
Prevent xstate to forward event to not existing playlist actor


prastoin:
How to reproduce the issue:
```
-Open Tab 1
-Tab 1 create a mpe room
-Open Tab 2
-Tab 1 add track
-Tab 2 displays an error as the related mpe room actor doesn't exist on Tab 2
```

We should use the ActorExists on every server events listeners

PR #391, opened 2022-01-14
prevent xstate to forward event to non existing playlist actor

prastoin:
close https://github.com/AdonisEnProvence/MusicRoom/issues/386


Devessier (approved):
Great!
We can merge.

Issue #392, opened 2022-01-14
Expand music player after creating or joining mtv room


prastoin:
after receiving a MTV_ROOM_SYNCED_CALLBACK or a MTV_ROOM_CREATE_CALLBACK, we should toggle on the music player fullscreen

this should also fix the following issue: https://github.com/AdonisEnProvence/MusicRoom/issues/375

prastoin:
Do we also want to expand the music player after joining a room ?

prastoin:
Client:
- [ ] Update AppMusicPlayerStateMachine to redirect view after create and join
- [ ] Udpate client integ tests
- [ ] Update e2e tests

Devessier:
This issue does not close https://github.com/AdonisEnProvence/MusicRoom/issues/375.

PR #393, opened 2022-01-17
Expand music player fullscreen after join and create

prastoin:
close https://github.com/AdonisEnProvence/MusicRoom/issues/392


prastoin:
wow this went wild, recreating a PR

Issue #387, opened 2022-01-14
Implement Refresh to MPE Library Search Engine


Devessier:
We want to be able to refresh the list of results.
Either by using Pull-to-refresh, or a button anywhere.

Devessier:
A cross-platform Pull-to-refresh could be easily implemented thanks to this library: https://github.com/NiciusB/react-native-web-refresh-control. [Open demo](https://snack.expo.dev/@niciusb/refreshcontrol-example)

PR #394, opened 2022-01-17
Finalize mpe search engine

Devessier:
Closes https://github.com/AdonisEnProvence/MusicRoom/issues/331
Closes https://github.com/AdonisEnProvence/MusicRoom/issues/332

PR #398, opened 2022-01-18
Add Pull to refresh to MPE Library

Devessier:
Closes https://github.com/AdonisEnProvence/MusicRoom/issues/387

Must be reviewed **after** https://github.com/AdonisEnProvence/MusicRoom/pull/394.

### DIFF
--- a/packages/client/global.d.ts
+++ b/packages/client/global.d.ts
@@ -1,1 +1,5 @@
 declare module 'xhr2';
+
+declare module 'react-native-web-refresh-control' {
+    export { RefreshControl } from 'react-native';
+}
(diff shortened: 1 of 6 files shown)


prastoin:
This is a great idea !
I was wondering if we would have to implem the refresh on every of our infinite loading search engine screen ?

prastoin (approved):
Huge job done
One question pending before mergin but 100% approved
Ready to be merged

Devessier:
I think this is not _necessary_ for other searches because, Library screen is mounted from the beginning of the application, and is never unmounted.
Other searches are meant to be used once, and are then unmounted. Therefore, on the next mounting, the results would have been cleared.

Although, this feature can be implemented for any other search. I just think it's not necessary for our current expectations.

prastoin:
I agree !

Issue #397, opened 2022-01-17
Redirect user on created mpe room view


prastoin:
After creating an mpe room, creator should be redirected to the created mpe room view screen
From now he's redirected on the library search engine

PR #401, opened 2022-01-19
after mpe room creation display created mpe room view

prastoin:
close https://github.com/AdonisEnProvence/MusicRoom/issues/397

### DIFF
--- a/packages/client/machines/appMusicPlaylistsMachine.ts
+++ b/packages/client/machines/appMusicPlaylistsMachine.ts
@@ -552,8 +552,7 @@ export function createAppMusicPlaylistsMachine({
                                             to: 'socketConnection',
                                         },
                                     ),
-
-                                    appMusicPlaylistsModel.actions.redirectToMpeLibrary(),
+                                    `redirectToMpeLibrary`,
                                 ],
                             },
                         },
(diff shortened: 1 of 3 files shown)


prastoin:
Should review and merge this https://github.com/AdonisEnProvence/MusicRoom/pull/404 before

packages/client/machines/appMusicPlaylistsMachine.ts
```diff
@@ -552,8 +552,7 @@ export function createAppMusicPlaylistsMachine({
                                             to: 'socketConnection',
                                         },
                                     ),
-
-                                    appMusicPlaylistsModel.actions.redirectToMpeLibrary(),
+                                    `redirectToMpeLibrary`,
```
prastoin, line 555: still redirecting to mpeLibrary after send req to the back to avoid rendering a white mpe room creation form confirmation screen

packages/client/machines/appMusicPlaylistsMachine.ts
```diff
@@ -573,14 +572,33 @@ export function createAppMusicPlaylistsMachine({
...
-                                        },
+                                actions: [
+                                    send((_, { state }) =>
+                                        appMusicPlaylistsModel.events.SPAWN_PLAYLIST_ACTOR_FROM_STATE(
+                                            {
+                                                roomID: state.roomID,
+                                                state,
+                                            },
+                                        ),
                                     ),
```
Devessier, line 583: I was wondering if we should not use [`raise`](https://xstate.js.org/docs/guides/actions.html#raise-action) action instead of [`send`](https://xstate.js.org/docs/guides/actions.html#send-action) when we want to do that kind of thing 🤔

I never used it but I will dive deeper to understand when we should favour one solution over the other.

Devessier (approved):
Nice work!
We can merge.

packages/client/machines/appMusicPlaylistsMachine.ts
prastoin: :thinking: you're right this is a bit a priority

PR #404, opened 2022-01-19
Render initially all tab initial screens

Devessier:


prastoin:
Did you know that an ant can lift up to 20 times it's body weight ?
Was thinking about that while reviewing this PR

prastoin (dismissed):
Great work ! ready and eager to merge this

Devessier:
Lazy Ant-Man!

prastoin:
Imo we need to update every integ and e2e client tests, now every selector might find several result for a query

prastoin:
We need to have a look on the assertions I've updated but also the assertion I had to remove, mainly only few `waitForElementToBeRemoved` assertions.
It happened that I've refactor those assertions expecting a number of matching elements.
Also still need to refactor e2e tests, checking tomorrow

packages/client/tests/tests-mpe-utils.ts
```diff
@@ -75,7 +75,9 @@ export async function createMpeRoom(): Promise<{
         1,
     );
 
-    const goToLibraryButton = screen.getByText(/library/i);
+    const goToLibraryButton = (
```
prastoin, line 78: I have no idea but this locator will fail during client tests in the ci, resolving to multiple elements :thinking: 
It's should not be possible to duplicate the bottomBar Icons, no issues in local

packages/client/__tests__/SearchTracks.test.tsx
```diff
@@ -247,10 +247,11 @@ test('SearchTracksScreen results are reset when pressing cancel button', async (
     const waitForTrackResultListItemToDisappearPromise =
         waitForElementToBeRemoved(() => screen.getByText(fakeTrack.title));
 
-    const cancelButton = screen.getByText(/cancel/i);
-    expect(cancelButton).toBeTruthy();
+    //Need to determine why this needs to be the first element ?
+    const firstCancelButton = screen.getAllByText(/cancel/i)[0];
+    expect(firstCancelButton).toBeTruthy();
```
Devessier, line 252: Is this because Search Tracks is before Library in the Bottom Bar?

packages/client/tests/tests-mpe-utils.ts
Devessier: Strange 😮

Devessier (commented):
That's a lot of chances in tests for a little change in our codebase!
I think we can merge.

I just want to point out that relying on the order of elements on screens is, to me, a weakness. We were doing it in our integration tests prior to this PR, and even in E2E tests.
I think that using a function such as `withinMpeRoomsLibraryScreen`, as we did in E2E tests, but in integration tests, to scope queries to a particular screen, would make our tests a lot more predictable, understandable and maintainable.

prastoin:
> That's a lot of chances in tests for a little change in our codebase! I think we can merge.
> 
> I just want to point out that relying on the order of elements on screens is, to me, a weakness. We were doing it in our integration tests prior to this PR, and even in E2E tests. I think that using a function such as `withinMpeRoomsLibraryScreen` in E2E tests, to scope queries to a particular screen, would make our tests a lot more predictable, understandable and maintainable.

I think you're completely right, this would be better in any form. But either retrieving the higher layer for our within assertion could be a pain :smile: This is clearly game changer for our test env

prastoin (approved):
reviewing with devessier agreement 
ready to be merged

PR #405, opened 2022-01-19
Refactor Bottom Tab Bar component

Devessier:
Instead of reimplementing this crucial component, we use and customize a component exported by React Navigation library.


prastoin (approved):
Indeed nice refactor :smile: 
Ready to be merged !

Issue #406, opened 2022-01-20
Mpe room view integration refactor


prastoin:
![image](https://user-images.githubusercontent.com/45004772/150374608-cc834e5d-b7cb-417c-9ade-67957fbb0e67.png)

Mainly we need to centralize the leave room and export as Mtv inside a bottom sheet modal as above, accessible from a settings button.
Also improve the global layout and responsiveness

Client:
- [x] Create a settings icon
- [x] On settings icon press display settings bottom sheet modal
- [x] In the settings bottom sheet modal put the export as mtv and leave mpe room button
- [x] Update integ tests to new layout
- [x] Update e2e tests to new layout

PR #407, opened 2022-01-21
Mpe room view refactor integration

prastoin:
close https://github.com/AdonisEnProvence/MusicRoom/issues/406

### DIFF
--- a/packages/client/e2e/_utils/mpe-e2e-utils.ts
+++ b/packages/client/e2e/_utils/mpe-e2e-utils.ts
@@ -1,6 +1,5 @@
 import { MpeChangeTrackOrderOperationToApply } from '@musicroom/types';
 import { expect, Page, Locator } from '@playwright/test';
-import { waitFor } from '@testing-library/react-native';
 import { random } from 'faker';
 import { assertIsNotNull } from '../_utils/assert';
 import { hitGoNextButton } from '../_utils/global';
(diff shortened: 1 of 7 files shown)


packages/client/e2e/_utils/mpe-e2e-utils.ts
```diff
@@ -320,6 +321,31 @@ export async function searchAndJoinMpeRoomFromMpeRoomsSearchEngine({
...
+
+    await waitFor(() => {
+        const exportButton = page.locator(
+            `css=[data-testid="export-mpe-to-mtv-button"]`,
+        );
+        expect(exportButton).toBeTruthy();
+        const leaveRoomButton = page.locator(
+            `css=[data-testid="leave-mpe-room-button"]`,
+        );
+        expect(leaveRoomButton).toBeTruthy();
```
Devessier, line 345: `toBeTruthy` assertion will do nothing in E2E tests. Actually, it will assert that `leaveRoomButton` is not falsy. Here, `leaveRoomButton` is an object of type `Locator`, and not the DOM elements pointed by the selector. We should be using `toBeVisible`, or any other assertion that will resolve the `Locator` into real elements.

Devessier: I continued to read all the files, and I understand why you wrote that. As our E2E tests and our integration tests have quite the same API, we can easily get lost between each other 😄

Devessier (changes requested):
Great job!
I like the UI!
I think we need to update slightly a utility function for E2E tests.

packages/client/e2e/_utils/mpe-e2e-utils.ts
prastoin: You're right mistake there, fixing it

Devessier (approved):
Nice!
We can merge.

PR #410, opened 2022-01-24
Fix test id conflict

### LINKED ISSUE #329: TestID conflict for mpe tracks cards same with summary and MTV tracks cards
We should have a container that static set a prefix to put on every testID that would be MTV or MPE

prastoin:
close #329

### DIFF
--- a/packages/client/tests/tests-utils.tsx
+++ b/packages/client/tests/tests-utils.tsx
@@ -10,6 +10,7 @@ import { DripsyProvider } from 'dripsy';
 import { datatype } from 'faker';
 import React from 'react';
 import { SafeAreaProvider } from 'react-native-safe-area-context';
+import invariant from 'tiny-invariant';
 import { AppContextProvider } from '../contexts/AppContext';
 import { SocketContextProvider } from '../contexts/SocketContext';
 import { useTheme } from '../hooks/useTheme';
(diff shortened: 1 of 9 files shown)


packages/client/tests/tests-utils.tsx
```diff
@@ -143,9 +143,20 @@ export function getFakeUsersList({
 }
 
 export function extractTrackIDFromCardContainerTestID(testID: string): string {
-    return testID.replace('-track-card-container', '');
+    return testID
+        .replace('-track-card-container', '')
+        .replace('mpe-', '')
+        .replace('mtv-', '');
```
Devessier, line 149: I would expect more defensive code here.
When I want to extract some data, I prefer to focus on what I want instead of what I don't want.

I would use a RegExp to extract the `trackID`, by hardly expecting the pattern. If the pattern changes, we would like other developers to check manually that this function still does what we want.

```suggestion
    const result = testID.match(/(?:mpe|mtv)-(?<trackID>.+)-track-card-container/);
    invariant(result !== null, 'Pattern was not recognized');
    
    const trackID = result.groups.trackID;
    invariant(typeof trackID === 'string', 'TrackID named group has not been captured');
    
    return trackID;
```

Devessier (approved):
Nice job!
We can merge.

I wrote some thoughts about `trackID` extraction from `testID`.

packages/client/tests/tests-utils.tsx
prastoin: You're right this clearly more readable for everyone.
Great suggestion thanks !
The testID prefix can be either 'mtv' or 'mpe' but also undefined
In this way we should add a third occurence the first regex query
```regex
(?:mpe-|mtv-|)(?<trackID>.+)-track-card-container
```
Pushing a related commit soon

Devessier (approved):
Nice!
We can merge.

PR #411, opened 2022-01-24
Make create room and join room states parallel

Devessier:
First step of https://github.com/AdonisEnProvence/MusicRoom/issues/375.

The status of connection to a room must be separated from room creation or joining. This is what this PR solves.


prastoin (approved):
Nice refactor !
Indeed we've been implemting this state machine as the  three main states were // but It wasn't really the case ahah
From there it will be really smooth to implem the export and display mtv room issue
Ready to be merged !

PR #412, opened 2022-01-25
Use date fns to display playlist total duration

prastoin:
close https://github.com/AdonisEnProvence/MusicRoom/issues/409


Devessier (approved):
Nice!
It's so good to delegate this kind of responsibility.
We can merge.

PR #415, opened 2022-01-25
Run E2E tests for every push to master

Devessier:


prastoin:
We've determined that we would not pursue any further this issue/PR

Issue #426, opened 2022-01-28
Get UserProfileInformation


prastoin:
We need to be able, after going in a user profile page, to fetch the related user profile page information to dynamically display the good cta nor information
Such as, viewing a following user profile page etc

Client:
- [x] create a `userProfileStateMachine` to manage the view, it will contain in it's context a `UserProfileInformation`
- [x] after display user profile page screen, we need to send a getUserProfileInformation http request to the back, and then merge the result to update the view, it should be handled inside the stateMachine
- [x] tests
- [x] create a my profile page screen, that hit the `my-profile-information` route
- [x] add an my profile icon CTA in the home top right header section
- [x] on my page page screen retrieve my profile information via the same `UserProfileInformation` state machine if possible ?
- [x] tests

Server:
- [x] create a post route `/user/profile-information` that takes as a body `userID: string` and `tmpAuthUserID: string` to be able to mock authentification from now
- [x] create a post route`/user/my-profile-information` that takes an `tmpAuthUserID: string` to be able to mock authentication from now
- [x] handle created route, if given user is found then return response with as body related `UserProfileInformation`
- [x] handle created route, if given user is not found then return an `404` res code
- [x] tests

Types:
- [x] create a `SharedUserProfileInformation` that contains for the moment `userID: string` `userNickname: string`
- [x] create a `UserProfileInformation` that extends `SharedUserProfileInformation`
- [x] create a `MyProfileInformation` that extends `SharedUserProfileInformation` adding a `devicesCounter: number`

prastoin:
`user/profile-information` route should not allow the request user to pass it's own userID
TODO
- [ ] prevent inside handler
- [ ] tests

PR #433, opened 2022-02-01
prevent get profile information requesting user retrieve his own information

prastoin:
related issue https://github.com/AdonisEnProvence/MusicRoom/issues/426


Devessier (approved):
Great small improvement.
We can merge.

PR #435, opened 2022-02-02
Get my profile information server

prastoin:
related to https://github.com/AdonisEnProvence/MusicRoom/issues/426

### DIFF
--- a/packages/server/tests/MyProfileInformation.test.ts
+++ b/packages/server/tests/MyProfileInformation.test.ts
@@ -1,0 +1,75 @@
+import Database from '@ioc:Adonis/Lucid/Database';
+import {
+    GetMyProfileInformationRequestBody,
+    GetMyProfileInformationResponseBody,
+} from '@musicroom/types';
+import User from 'App/Models/User';
+import { datatype, internet } from 'faker';
+import test from 'japa';
+import supertest from 'supertest';
+import urlcat from 'urlcat';
+import { USER_ROUTES_GROUP_PREFIX } from '../start/routes';
+import { BASE_URL, initTestUtils } from './utils/TestUtils';
+
+test.group('Users Profile information tests', (group) => {
+    const { createUserAndGetSocket, createSocketConnection } = initTestUtils();
+
+    group.beforeEach(async () => {
+        await Database.beginGlobalTransaction();
+    });
+
+    group.afterEach(async () => {
+        await Database.rollbackGlobalTransaction();
+    });
+
+    test('It should retrieve my profile information', async (assert) => {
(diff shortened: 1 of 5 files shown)


packages/server/tests/MyProfileInformation.test.ts
```diff
@@ -0,0 +1,75 @@
...
+test.group('Users Profile information tests', (group) => {
+    const { createUserAndGetSocket, createSocketConnection } = initTestUtils();
+
+    group.beforeEach(async () => {
+        await Database.beginGlobalTransaction();
+    });
+
+    group.afterEach(async () => {
+        await Database.rollbackGlobalTransaction();
+    });
```
Devessier, line 23: If you create sockets in tests, I think you should ensure they are all cleared before running next tests.

```suggestion
    group.beforeEach(async () => {
        initSocketConnection();
        await Database.beginGlobalTransaction();
    });

    group.afterEach(async () => {
        await disconnectEveryRemainingSocketConnection();
        sinon.restore();
        await Database.rollbackGlobalTransaction();
    });
```

packages/server/tests/UserProfileInformation.test.ts
```diff
@@ -31,7 +33,7 @@ test.group('Users Profile information tests', (group) => {
         });
 
         const { body: rawBody } = await supertest(BASE_URL)
-            .post('/user/profile-information')
+            .post(urlcat(USER_ROUTES_GROUP_PREFIX, 'profile-information'))
```
Devessier, line 36: I do not really have a strong opinion on this, but using `USER_ROUTES_GROUP_PREFIX` here breaks black-box testing principles.

Devessier (changes requested):
Great work!
I will be able to add visibility settings on the types you created!

I think we should update `beforeEach` and `afterEach` in the test file you created to do not encounter bigger issues later.

packages/server/tests/MyProfileInformation.test.ts
prastoin: Yes you're right ! forgot to add required test socket life cycle method here
thanks

packages/server/tests/UserProfileInformation.test.ts
prastoin: Yes you're right :thinking: 
We were already doing the same for both MPE and MTV PREFIX inside others tests
I don't really know if we came to a conclusion on them ?
We could duplicate them inside tests utils ?
Gonna forward an implem suggestion

Devessier (approved):
Nice! Black-box testing is not broken 👌
We can merge.

PR #437, opened 2022-02-02
Get my profile information client

### LINKED ISSUE #426: Get UserProfileInformation

prastoin:
close #426 

to be merged before : https://github.com/AdonisEnProvence/MusicRoom/pull/435


packages/client/screens/MyProfile/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+import { getFakeUserID } from '../../contexts/SocketContext';
+import { createMyProfileInformationMachine } from '../../machines/myProfileInformationMachine';
+import { MusicPlaylistEditorUsersSearchNavigator } from '../../navigation';
+
+const MyProfileScreen: React.FC<UserProfileScreenProps> = ({ navigation }) => {
+    const insets = useSafeAreaInsets();
+    const userID = getFakeUserID();
+
+    const userProfileInformationMachineConfigured = useMemo(
+        createMyProfileInformationMachine,
```
Devessier, line 21: It does the same thing but I think people would be more familiar with this syntax:

```suggestion
        () => createMyProfileInformationMachine(),
```

packages/client/screens/MyProfile/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+        [],
+    );
+
+    const userProfileInformationService = useInterpret(
+        userProfileInformationMachineConfigured,
+    );
```
Devessier, line 27: Another pattern is possible, but we have never tried it:

```suggestion
    const userProfileInformationService = useInterpret(
        () => createMyProfileInformationMachine(),
    );
```

Devessier (dismissed):
Really nice job!
We can merge.

packages/client/screens/MyProfile/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+    const userID = getFakeUserID();
+
+    const userProfileInformationMachineConfigured = useMemo(
+        () => createMyProfileInformationMachine(),
+        [],
+    );
+
+    const userProfileInformationService = useInterpret(
+        () => userProfileInformationMachineConfigured,
+    );
```
Devessier, line 27: Using `useMemo` to compute a _state machine function creator_ once does the same thing as providing a function to `useInterpret`.

In my previous comments, I meant more something like that:

```suggestion
    const userProfileInformationService = useInterpret(
        () => createMyProfileInformationMachine(),
    );
```

packages/client/screens/UserProfile/index.tsx
```diff
@@ -25,13 +24,12 @@ const UserProfileScreen: React.FC<UserProfileScreenProps> = ({
...
                 userID,
-                config: getUserProfileInformationMachineOptions(),
             }),
         [userID],
     );
 
     const userProfileInformationService = useInterpret(
-        userProfileInformationMachineConfigured,
+        () => userProfileInformationMachineConfigured,
     );
```
Devessier, line 33: As for `packages/client/screens/MyProfile/index.tsx` file, providing a function to `useInterpret` serves the same purpose as using a `useMemo` to compute the state machine only once.

```suggestion
    const userProfileInformationService = useInterpret(
        () => createUserProfileInformationMachine({ userID }),
    );
```

Devessier (approved):
Nice! I wrote some more comments to precise my previous thoughts.
We can merge.

packages/client/screens/MyProfile/index.tsx
prastoin: Ok I see, only because `createMyProfileInformationMachine` does not take args right ?

packages/client/screens/UserProfile/index.tsx
prastoin: Isn't it a problem for the userID to do not be watched in the useMemo ?

packages/client/screens/SearchTrackScreen.tsx
```diff
@@ -122,39 +122,35 @@ const SearchTrackScreen: React.FC<SearchTabSearchTracksScreenProps> = ({
...
+                    initialTracksIDs: [selectedTrackID],
+                });
             },
-        });
-    }, [appMusicPlaylistsActorRef, sendToMusicPlayerMachine]);
-    const searchTracksScreenService = useInterpret(
-        searchTrackScreenMachineConfigured,
-    );
+        },
+    });
```
Devessier, line 153: I like it!

packages/client/screens/UserProfile/index.tsx
```diff
@@ -1,37 +1,27 @@
...
-                config: getUserProfileInformationMachineOptions(),
-            }),
-        [userID],
-    );
 
-    const userProfileInformationService = useInterpret(
-        userProfileInformationMachineConfigured,
+    const userProfileInformationService = useInterpret(() =>
+        createUserProfileInformationMachine({ userID }),
     );
```
Devessier, line 25: Yeah babe!

packages/client/screens/UserProfile/index.tsx
```diff
@@ -106,4 +96,27 @@ const UserProfileScreen: React.FC<UserProfileScreenProps> = ({
...
+    if (userID === undefined || userID === null) {
+        throw new Error('UserProfile received no userID');
+    }
+
+    return (
+        <UserProfileContent
+            route={route}
+            navigation={navigation}
+            userID={userID}
+            key={userID}
```
Devessier, line 117: This is where things get interesting. 👀

Devessier (approved):
Really nice!
We can merge.

Issue #439, opened 2022-02-03
UserProfile and myProfile sections ( playlist following follower )


prastoin:
Client:
- [x] my profile display the Playlist Counter, on press redirect to my library screen
- [x] my profile display the Follower Counter, on press redirect to my follower search engine
- [x] my profile display the Following Counter, on press redirect to my following search engine
- [x] tests
- [x] user profile display Playlist depending on visibility, on press redirect to other user mpe search engine
- [x] user profile display follower depending on visibility, on press redirect to user follower search engine
- [x] user profile display following depending on visibility, on press redirect to user following search engine
- [x] tests

server:
- [x] refactor `user/get-profile-information` route to compute exported props depending on related fields visibility
- [x] tests
- [x] refactor `user/my-profile-information` route to compute new `MyProfileInformation` props
- [x] tests

types:
- [x] On UserProfileInformation add the following fields:
```ts
playlistCounter: number | undefined
followerCounter: number | undefined
followingCounter: number | undefined
```
- [x] On MyProfileInformation add the following fields:
```ts
playlistCounter: number
followerCounter: number
followingCounter: number
```

PR #440, opened 2022-02-04
Get user profile information apply visibility server

prastoin:
related to https://github.com/AdonisEnProvence/MusicRoom/issues/439

### DIFF
--- a/packages/server/app/Controllers/Http/SearchUsersController.ts
+++ b/packages/server/app/Controllers/Http/SearchUsersController.ts
@@ -6,12 +6,41 @@ import {
     GetUserProfileInformationResponseBody,
     SearchUsersRequestBody,
     SearchUsersResponseBody,
+    UserSettingVisibility,
 } from '@musicroom/types';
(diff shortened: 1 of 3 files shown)


packages/server/app/Controllers/Http/SearchUsersController.ts
```diff
@@ -6,13 +6,43 @@ import {
...
+        }
+        case UserSettingVisibility.Values.FOLLOWERS_ONLY: {
+            if (requestingUserIsfollowingRelatedUser) {
+                return fieldValue;
+            }
+            return undefined;
+        }
+        default: {
+            console.error(`unknown switch ue case ${fieldVisibility}`);
+            return undefined;
```
Devessier, line 41: I like to throw an error when an unreachable state is reached, because it means the program can not continue to work properly. Therefore, the best thing to do is to stop it as soon as possible 😆

Devessier (approved):
Nice work!
We can merge.

Devessier (approved):
Nice, we can merge.

Issue #423, opened 2022-01-28
Follow server


prastoin:
In this issue will be treated the follow user server implementation

**Todo**

server:

Logic/
- [x] create a `post` http route `user/follow` that takes a `userID` in body **and a tmpAuthUserID** to be able to mock auth for the moment
- [x] handle the created route, add to the related `tmpAuthUserID` following list the given `userID`
- [x] ?? if given followed user is already in the following list throw an error ??
- [x] if the operation is a success send back an 200 response and compute a new `UserProfileInformation` to send back as response body, do not forget to maintain the new added `followingCounter` & `followerCounter` freshly added to the `UserProfileInformation` type

Data set up/
- [x] update get user profile information route to compute added props to `userProfileInformation`
- [x] Create a new pivot table `following_user_followed_user` with column: `following_user_uuid` and `followed_user_uuid` that refers to user table
- [x] inside the User model add a new `following` `ManyToMany` relationship toward other `User` model
- [x] inside the user model add a new `followers` `ManyToMany` relationship toward other `User` model

Both following and follower User model props `ManyToMany` configuration should look something like this:
```ts
    @manyToMany(() => User, {
        localKey: 'uuid', //following
        relatedKey: 'uuid', //followed

        pivotTable: 'following_user_followed_user',
        pivotForeignKey: 'following_user_uuid',
        pivotRelatedForeignKey: 'followed_user_uuid',
    })
    public following: ManyToMany<typeof User>;
    
    @manyToMany(() => User, {
        localKey: 'uuid', //followed
        relatedKey: 'uuid', //following

        pivotTable: 'following_user_followed_user',
        pivotForeignKey: 'followed_user_uuid',
        pivotRelatedForeignKey: 'following_user_uuid',
    })
   public follower: ManyToMany<typeof User>;
 ```

Note: https://github.com/AdonisEnProvence/MusicRoom/issues/426 should have been close before this issue
types:
- [x] add to `UserProfileInformation` a `followerCounter: number` and a `followingCounter: number`
- [x] add `following: boolean` to `UserProfileInformation`

PR #444, opened 2022-02-07
Follow feature server

prastoin:
close https://github.com/AdonisEnProvence/MusicRoom/issues/423

### DIFF
--- a/packages/server/database/migrations/1644229555565_following_user_followed_user.ts
+++ b/packages/server/database/migrations/1644229555565_following_user_followed_user.ts
@@ -1,0 +1,30 @@
+import BaseSchema from '@ioc:Adonis/Lucid/Schema';
+
+export default class FollowingUserFollowedUser extends BaseSchema {
+    protected tableName = 'following_user_followed_user';
+
+    //pivot table
+    public up(): void {
+        this.schema.createTable(this.tableName, (table) => {
+            table.increments('id');
+            table
+                .uuid('following_user_uuid')
+                .references('users.uuid')
+                .onDelete('CASCADE');
+            table
+                .uuid('followed_user_uuid')
+                .references('users.uuid')
+                .onDelete('CASCADE');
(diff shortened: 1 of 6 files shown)


packages/server/database/migrations/1644229555565_following_user_followed_user.ts
```diff
@@ -0,0 +1,30 @@
+import BaseSchema from '@ioc:Adonis/Lucid/Schema';
+
+export default class FollowingUserFollowedUser extends BaseSchema {
+    protected tableName = 'following_user_followed_user';
+
+    //pivot table
+    public up(): void {
+        this.schema.createTable(this.tableName, (table) => {
+            table.increments('id');
```
Devessier, line 9: Do you think we should use a UUID auto-generated instead of an integer?

Devessier (approved):
Really nice job! This feature will be so cool.
We can merge.

packages/server/database/migrations/1644229555565_following_user_followed_user.ts
prastoin: I understand and share your concerns, it might be more logical and idiomatic to use uuid here too.
I've been using an integer as primary and ID here because we've also been doing so for the `mpeRoom` pivot table.
Can't remember if this was a defined choice or just a forgotten detail ? 
This might also be a way to differentiate pivot from classic tables

Devessier: Currently when we had a table with an UUID as the primary key, we were generating the UUID in JS code. The issue here is that I don't think we would be able to create the primary key like that.
As a consequence, we would need to use a function such as `uuid_generate_v4` (https://www.postgresql.org/docs/current/uuid-ossp.html) in the `DEFAULT` case of the primary key column. This would introduce a dependency to PostgreSQL as I do not think this function is available in other databases.

Using a `SERIAL` here, as you do, seems a lot easier 😄

PR #447, opened 2022-02-07
Plug profile information follow counters

prastoin:


packages/server/app/Controllers/Http/SearchUsersController.ts
```diff
@@ -87,14 +87,19 @@ export default class SearchUsersController {
...
+        });
+        const requestingUserIsfollowingRelatedUser =
+            requestingUser.following.length > 0;
 
         const relateduser = await User.findOrFail(userID);
         await relateduser.load('playlistsVisibilitySetting');
         await relateduser.load('relationsVisibilitySetting');
         await relateduser.load('mpeRooms');
+        await relateduser.load('following');
+        await relateduser.load('followers');
```
Devessier, line 102: If we only want the count of following or followers, maybe could we use [`relatedUser.loadCount()`](https://docs.adonisjs.com/guides/models/relationships#lazy-load-relationship-aggregates) to do not load the information of the users?

packages/server/app/Controllers/Http/SearchUsersController.ts
```diff
@@ -144,17 +149,17 @@ export default class SearchUsersController {
         const user = await User.findOrFail(tmpAuthUserID);
         await user.load('devices');
         await user.load('mpeRooms');
-        const followingCounter = 12; //FIXME follow unfollow dev
-        const followersCounter = 13; //FIXME follow unfollow dev
+        await user.load('followers');
+        await user.load('following');
```
Devessier, line 153: Same comment about the user of [`user.loadCount()`](https://docs.adonisjs.com/guides/models/relationships#lazy-load-relationship-aggregates).

Devessier (approved):
Nice job!
Lucid is a really elegant ORM.
I suggest a modification, that simplifies code and reduces data amount loaded from the database.
We can merge.

packages/server/app/Controllers/Http/SearchUsersController.ts
```diff
@@ -95,35 +95,38 @@ export default class SearchUsersController {
...
         const {
             nickname: userNickname,
             playlistsVisibilitySetting,
             relationsVisibilitySetting,
         } = relateduser;
 
+        await relateduser
+            .loadCount('following')
+            .loadCount('followers')
+            .loadCount('mpeRooms');
```
Devessier, line 111: Really nice!

packages/server/app/Controllers/Http/SearchUsersController.ts
```diff
@@ -95,35 +95,38 @@ export default class SearchUsersController {
...
+
         const playlistsCounter =
             getUserProfileInformationDependingOnItsVisibility({
-                fieldValue: relateduser.mpeRooms.length,
+                fieldValue: parseInt(relateduser.$extras.mpeRooms_count),
```
Devessier, line 115: It seems that [the radix is not necessarily 10](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt#parameters) and needs to be provided to `parseInt` explicitly.
Usually I use `Number()` instead of `parseInt()` because `parseInt()` [parses the input](https://thisthat.dev/number-constructor-vs-parse-int/), while `Number()` coerces its type.

packages/server/app/Controllers/Http/SearchUsersController.ts
```diff
@@ -147,20 +150,25 @@ export default class SearchUsersController {
...
-        await user.load('followers');
-        await user.load('following');
+        const { nickname: userNickname } = user;
+
+        //After this user model column are erased
+        await user
+            .loadCount('following')
+            .loadCount('followers')
+            .loadCount('mpeRooms')
+            .loadCount('devices');
```
Devessier, line 160: Do you think Lucid will do only one request for these four aggregates? It would be super fun 😄

Devessier (approved):
Nice job! We discover new features of Lucid every day!
We can merge.

packages/server/app/Controllers/Http/SearchUsersController.ts
prastoin: I think because if you don't regroup them in one, the last will erased previous `$extras` instance

packages/server/app/Controllers/Http/SearchUsersController.ts
prastoin: updating

Devessier (approved):
Perfect.
We can merge.

Issue #430, opened 2022-01-31
my | other follower & my | other following search engine Server


prastoin:
Server:
- [x] Create a new http controller `UserRelationController`
This class will contains at least two method:
1/ `UserFollowerSearch` taking a `userID` returns a `UserSummary[]` with pagination
2/ `UserFollowingSearch` taking a `userID` returns a `UserSummary[]` with pagination
- [x] create a `search/user-follower` taking a `userID` in req body and needed pagination, the route will return a `UserSummary[]` + pagination in response body.
This route will triggers `UserFollowerSearch` passing the given `userID`
- [x] create a `search/user-following` taking a `userID` and needed pagination in req body, the route will return a `UserSummary[]` + pagination in response body.
This route will triggers a  `UserFollowingSearch` passing the given `userID`
- [x] create a `search/my-follower` route taking required pagination in body the route returns a `UserSummary[]` + pagination in response body.
This route will triggers a `UserFollowerSearch` passing retrieved `userID` from auth
- [x] create a `search/my-following` route taking required pagination, the route returns a `UserSummary[]` + pagination in response body.
This route will triggers a `UserFollowingSearch` passing retrieved `userID` from auth
- [x] on each routes check related user relations settings visibility to return or not data
- [x] tests

PR #461, opened 2022-02-11
My other followers following search engine server

prastoin:
related to https://github.com/AdonisEnProvence/MusicRoom/issues/430

### DIFF
--- a/packages/server/app/Controllers/Http/SearchUsersController.ts
+++ b/packages/server/app/Controllers/Http/SearchUsersController.ts
@@ -271,6 +271,7 @@ export default class SearchUsersController {
         const { tmpAuthUserID, userID } =
             GetUserProfileInformationRequestBody.parse(rawBody);
 
+        await User.findOrFail(tmpAuthUserID);
         const userProfileInformation = await requestUserProfileInformation({
             requestingUserID: tmpAuthUserID,
             userID,
--- a/packages/server/app/Controllers/Http/SearchUsersController.ts
+++ b/packages/server/app/Controllers/Http/SearchUsersController.ts
@@ -5,6 +5,10 @@ import {
     GetMyProfileInformationResponseBody,
     GetUserProfileInformationRequestBody,
     GetUserProfileInformationResponseBody,
+    ListUserFollowersRequestBody,
+    ListUserFollowersResponseBody,
+    ListUserFollowingRequestBody,
+    ListUserFollowingResponseBody,
     SearchUsersRequestBody,
     SearchUsersResponseBody,
     UnfollowUserRequestBody,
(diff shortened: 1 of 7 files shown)


packages/server/app/Controllers/Http/SearchUsersController.ts
```diff
@@ -138,6 +186,83 @@ export default class SearchUsersController {
...
+            {
+                relatedUserID: userID,
+                requestingUserID: tmpAuthUserID,
+            },
+        );
+        //
+        const usersFollowersPagination = await User.query()
+            .whereNot('uuid', userID)
+            .andWhere('nickname', 'ilike', `${searchQuery}%`)
+            .andWhereHas('following', (userQuery) => {
```
Devessier, line 207: **👏 praise:** ‏To get a user's followers, you get all the users that follow her and are not her. One's brain must be active to understand this relation.

packages/server/tests/ListUserFollowers.test.ts
```diff
@@ -0,0 +1,385 @@
...
+        const searchedUser = await User.create({
+            uuid: searchedUserID,
+            nickname: internet.userName(),
+        });
+        await User.create({
+            uuid: requestingUserID,
+            nickname: internet.userName(),
+        });
+
+        const searchQuery = datatype.string(1);
```
Devessier, line 171: **👏 praise:** ‏Nice!

packages/server/tests/ListUserFollowers.test.ts
```diff
@@ -0,0 +1,385 @@
...
+        ).filter(
+            (el: UserSummary | undefined): el is UserSummary =>
+                el !== undefined,
+        );
+        ///
+
+        //filtering options
+        const matchingSearchQuerySortedByNicknameSearchedUserFollowers =
+            searchedUserFollowersUserSummary
+                .filter((user) => user.nickname.startsWith(searchQuery))
```
Devessier, line 213: **❔ question:** ‏‏In the database we filter users with `ilike` operator. I think that the `filter()` method here should be case-insensitive too.

packages/server/tests/ListUserFollowers.test.ts
```diff
@@ -0,0 +1,385 @@
...
+        await User.create({
+            uuid: requestingUserID,
+            nickname: internet.userName(),
+        });
+        const searchedUser = await User.create({
+            uuid: searchedUserID,
+            nickname: internet.userName(),
+        });
+
+        const privateVisibility = await getVisibilityDatabaseEntry('PRIVATE');
```
Devessier, line 308: **❔ question:** ‏Why did not you use `UserSettingVisibility.enum.PRIVATE` here?

Devessier (dismissed):
You did an impressive work on server tests! Great job.
We can merge.

packages/server/tests/ListUserFollowers.test.ts
prastoin: Indeed you're right, but isn't it not required as I manually prepend `searchQuery ` at the nicknames beginning ? :thinking:

packages/server/tests/ListUserFollowers.test.ts
prastoin: Oh forgot to comment on the PR about it
For your beautiful eyes of course ;) jking
Wanted to try it out

packages/server/tests/ListUserFollowers.test.ts
prastoin: So if I understand correctly, you would see something either like:
```ts
        //filtering options
        const matchingSearchQuerySortedByNicknameSearchedUserFollowers =
            searchedUserFollowersUserSummary
                .filter((user) =>
                    user.nickname.toLowerCase().startsWith(searchQuery),
                )
                .sort((a, b) => {
                    if (a.nickname.toLowerCase() < b.nickname.toLowerCase()) {
                        return -1;
                    }
                    if (a.nickname.toLowerCase() > b.nickname.toLowerCase()) {
                        return 1;
                    }
                    return 0;
                });
        ///
```

nor

```ts
        //filtering options
        const matchingSearchQuerySearchedUserFollowersUserSummary =
            searchedUserFollowersUserSummary.filter((user) =>
                user.nickname.toLowerCase().startsWith(searchQuery),
            );
        const matchingSearchQuerySortedByNicknameSearchedUserFollowers =
            matchingSearchQuerySearchedUserFollowersUserSummary.sort((a, b) => {
                if (a.nickname.toLowerCase() < b.nickname.toLowerCase()) {
                    return -1;
                }
                if (a.nickname.toLowerCase() > b.nickname.toLowerCase()) {
                    return 1;
                }
                return 0;
            });
        ///
```

Devessier: Something like that:

```ts
        const matchingSearchQuerySortedByNicknameSearchedUserFollowers =
            searchedUserFollowersUserSummary
                .filter((user) =>
                    // both `user.nickname` and `searchQuery` should be lower-cased
                    user.nickname.toLowerCase().startsWith(searchQuery.toLowerCase()),
                )
                .sort((a, b) => {
                    if (a.nickname.toLowerCase() < b.nickname.toLowerCase()) {
                        return -1;
                    }
                    if (a.nickname.toLowerCase() > b.nickname.toLowerCase()) {
                        return 1;
                    }
                    return 0;
                });
```

prastoin: Ok I see makes sense !
Pushing an update

prastoin: Wondering if on mac OS those tests would fail :open_mouth: 
Could you run them? ( [related article](https://dba.stackexchange.com/questions/106964/why-is-my-postgresql-order-by-case-insensitive/131471#131471) )

Devessier: I'll try now. I think it won't fail as our PostgreSQL database is run inside a Docker container, which is presumably based on Linux 😮.
Interesting question!

Devessier: The tests succeeded. 👌

Devessier (approved):
Nice job!
We can merge.

Issue #354, opened 2021-12-20
Other Users MPE Search Engine - Server


Devessier:
## Todo

- [x] Create method `fetchUserRooms` on `MpeSearchRoomsHttpController`
- [x] Create route `/mpe/search-user-rooms` that uses `fetchUserRooms` method that takes a `userID` and a `tmpAuthUserID` as a body
- [x] This method returns all the **public** rooms of the given user if the correct user profile settings are set, precisely the `playtlistVisibility` value
- [x] if related user `playtlistVisibility` is available for the requesting user then send back the results else throw a 403 error
- [x] This method, if receiving a `searchQuery`, filters rooms by it
- [x] Pagination

prastoin:
We should keep in mind that later when we will be working on the auth features, a user will be able to set as private this information

Devessier:
Right.
We could do two things:

- When we fetch the information of a user, we get a flag indicating whether MPE rooms list is private or public
- In `fetchUserRooms` method we check for this authorization before making MPE rooms searching query

PR #463, opened 2022-02-11
Create route to return other user's MPE rooms

Devessier:
Closes https://github.com/AdonisEnProvence/MusicRoom/issues/354


packages/server/app/Controllers/Http/UsersController.ts
```diff
@@ -0,0 +1,98 @@
...
+import User from 'App/Models/User';
+import UserService from 'App/Services/UserService';
+import { fromMpeRoomsToMpeRoomSummaries } from '../Ws/MpeRoomsWsController';
+
+interface ComputeUserCanQueryOtherUserMpeRoomsArgs {
+    user: User;
+    queriedUser: User;
+}
+
+async function getIfUserCanQueryOtherUserMpeRooms({
```
prastoin, line 16: **praise:** looks great !

packages/server/tests/MpeSearchEngine.test.ts
```diff
@@ -281,6 +285,298 @@ test.group('My MPE Rooms Search', (group) => {
...
+        await Database.beginGlobalTransaction();
+    });
+
+    group.afterEach(async () => {
+        await disconnectEveryRemainingSocketConnection();
+        sinon.restore();
+        await Database.rollbackGlobalTransaction();
+    });
+
+    test("It should send back all other user's public MPE rooms", async (assert) => {
```
prastoin, line 306: **Suggestion:** Unless I'm mistaken, the related user playlistVisibility is not covered here ?
In other words the method should be tested here `getIfUserCanQueryOtherUserMpeRooms` ?

packages/server/app/Controllers/Http/UsersController.ts
```diff
@@ -0,0 +1,98 @@
+import { HttpContextContract } from '@ioc:Adonis/Core/HttpContext';
```
prastoin, line 1: In the last [last merged PR](https://github.com/AdonisEnProvence/MusicRoom/commit/9b8075fcd126181f126cf397e4fc5770f5b58193), we've been creating new controller especially the `UserProfileController` I don't know if we wanna merge both in one ?

Devessier: We can, indeed!
I created this controller before we merged your PR.

packages/server/tests/MpeSearchEngine.test.ts
Devessier: You're right, I'll add a test to ensure an error is thrown when querying the rooms of a user that disallows it (either by `PRIVATE` OR `FOLLOWERS_ONLY` prop).

packages/server/app/Controllers/Http/UsersController.ts
prastoin: Great !

Issue #442, opened 2022-02-04
From any user search engine, if use is pressing his card he should be redirected to myProfileInformationScreen


prastoin:

prastoin:
We determined that this use case will occurs only on the MtvUsersList screen as it's the only one that displays user summary list that could contain the requesting user, user card
We have to determine what to do when the user is hitting his own user card, nothing ? redirect ?

PR #469, opened 2022-02-18
mtv users list on my user card press redirect to my profile

prastoin:
close https://github.com/AdonisEnProvence/MusicRoom/issues/442


Devessier (approved):
Nice job. We can merge.

Issue #429, opened 2022-01-31
my | other follower & my | other following search engine Client


prastoin:
Both my follower, my following, other user follower, other user following search route will send back the same data type, a userSummary[]
In this way we should centralize this logic inside the same machine, as we could have been doing for the mpeSearchEngine in the universalMpeSearchEngineMachine

Todo:
- [x] Create a  machine universalSearchEngineMachine to centralize every search engine logic, ofc pass a fetch method to the machine to hit the corresponding route. ?? Or use existing machine ??
- [x] Create a my follower search engine screen
- [x] Create a my following search engine screen
- [x] Create a other user or followed user following search engine screen
- [x] Create a other user or followed user follower search engine screen
- [x] Add to my user profile page, the following counter redirecting to my following search engine screen
- [x] Add to my user profile page, the followers counter redirecting to my followers search engine screen
- [x] Add to other or following user profile page, the followers counter redirecting to other user followers search engine screen
- [x] Add to other or following user profile page, the following counter redirecting to other user followers search engine screen

Issue #428, opened 2022-01-31
My & other following & my & other followers user search engine - PLAN


prastoin:
New user search engine are added during this feature implem.
We wanna retrieve every following but also follower client owner information
Both my follower, my following, other user follower, other user following search route will send back the same data type, a userSummary[]
In this way we should centralize this logic inside the same machine, as we could have been doing for the mpeSearchEngine in the universalMpeSearchEngineMachine

TODO
- [x] #429 
- [x] #430

PR #471, opened 2022-02-28
My followers following search screens

prastoin:
close https://github.com/AdonisEnProvence/MusicRoom/issues/428
close https://github.com/AdonisEnProvence/MusicRoom/issues/429

### DIFF
--- a/packages/client/screens/MyProfile/MyFollowers/index.tsx
+++ b/packages/client/screens/MyProfile/MyFollowers/index.tsx
@@ -5,7 +5,6 @@ import { ActorRef } from 'xstate';
 import { useActor, useMachine } from '@xstate/react';
 import { FlatList, TouchableOpacity } from 'react-native';
 import { RefreshControl } from 'react-native-web-refresh-control';
-import invariant from 'tiny-invariant';
 import { MyFollowersScreenProps } from '../../../types';
 import { AppScreenWithSearchBar } from '../../../components/kit';
 import {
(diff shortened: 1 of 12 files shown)


packages/client/screens/MyProfile/MyFollowers/index.tsx
```diff
@@ -0,0 +1,161 @@
...
+const MyFollowersSearchScreen: React.FC<MyFollowersScreenProps> = ({
+    navigation,
+}) => {
+    const insets = useSafeAreaInsets();
+    const sx = useSx();
+    const [screenOffsetY, setScreenOffsetY] = useState(0);
+    const initialNumberOfItemsToRender = IS_TEST ? Infinity : 10;
+
+    const [userFollowersSearchState, userFollowersSearchMachineSend] =
+        useMachine(() => myFollowerSearchMachine);
```
Devessier, line 27: **💭 thought:** ‏As `myFollowerSearchMachine` remains always the same, it's not necessary to use a function here.

packages/client/screens/MyProfile/MyFollowing/index.tsx
```diff
@@ -0,0 +1,161 @@
...
+const MyFollowingSearchScreen: React.FC<MyFollowingScreenProps> = ({
+    navigation,
+}) => {
+    const insets = useSafeAreaInsets();
+    const sx = useSx();
+    const [screenOffsetY, setScreenOffsetY] = useState(0);
+    const initialNumberOfItemsToRender = IS_TEST ? Infinity : 10;
+
+    const [userFollowingSearchState, userFollowingSearchMachineSend] =
+        useMachine(() => myFollowingSearchMachine);
```
Devessier, line 27: **💭 thought:** ‏As `myFollowingSearchMachine` always remains the same, it's not necessary to use a function here.

packages/client/screens/MyProfile/MyFollowing/index.tsx
```diff
@@ -0,0 +1,161 @@
...
+                                              }}
+                                          >
+                                              Load more
+                                          </Text>
+                                      </TouchableOpacity>
+                                  </View>
+                              );
+                          }
+                        : undefined
+                }
```
Devessier, line 155: **💭 thought:** ‏I recently discovered that `ListFooterComponent` prop accepts either a React function (as here) or a React component. The advantage of using a React component here is that React would be able to cache the Footer component. As it is written currently, a new component is created for each render, and I think that React is unmounting it for each render. This is an interesting thing that I want to deepen 👀.

See [React Native documentation for FlatList as reference](https://reactnative.dev/docs/flatlist#listfootercomponent)

Devessier (dismissed):
Nice job!
I like that you think your tests in term of _models_. 👍 
I wrote some thoughts, we can merge.

packages/client/screens/MyProfile/MyFollowing/index.tsx
prastoin: Interesting :open_mouth:

Devessier (approved):
Great job!
We can merge.

PR #472, opened 2022-02-28
Solve server tests logs on CI issue

prastoin:
We have an issue with the server tests logs inside the CI when it's failing
In an attempt to solve this problem we have decided to set up a stop tests on first fail policy
In this way maybe we will be able to retrieve the failing test context inside the ci logs


prastoin:
This is not changing anything !
About to close this issue

prastoin:
Tan tan dam plot twist, lerna was the imposter !

Devessier (approved):
Perfect!
We can merge.

Issue #465, opened 2022-02-14
MyProfile screen handle playlist counter on press


prastoin:
By pressing on the my profile playlist counter it should redirect the user to the my library screen

- [x] update onPress function
- [x] tests

PR #473, opened 2022-02-28
Handle my profile playlist counter

prastoin:
close https://github.com/AdonisEnProvence/MusicRoom/issues/465

### DIFF
--- a/packages/client/screens/MyProfile/__tests__/MyProfile.test.tsx
+++ b/packages/client/screens/MyProfile/__tests__/MyProfile.test.tsx
@@ -71,3 +71,55 @@ test('It should display my profile not found screen', async () => {
         });
     });
 });
+
+test('It should display my library after pressing playlists counter', async () => {
+    const userID = testGetFakeUserID();
+
+    db.myProfileInformation.create({
+        userID,
+        devicesCounter: 3,
+        playlistsCounter: 4,
+        followersCounter: 5,
+        followingCounter: 6,
+        userNickname: internet.userName(),
+    });
+
+    const screen = await renderApp();
+
+    expect(screen.getAllByText(/home/i).length).toBeGreaterThanOrEqual(1);
+
+    const myProfileButton = screen.getByTestId('open-my-profile-page-button');
+    expect(myProfileButton).toBeTruthy();
+
+    fireEvent.press(myProfileButton);
+
+    const playlistsCounter = await waitFor(() => {
+        expect(screen.getByTestId('my-profile-page-container')).toBeTruthy();
+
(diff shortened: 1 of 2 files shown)


packages/client/screens/MyProfile/__tests__/MyProfile.test.tsx
```diff
@@ -71,3 +71,54 @@ test('It should display my profile not found screen', async () => {
...
+        return playlistsCounter;
+    });
+
+    expect(playlistsCounter).toBeTruthy();
+
+    fireEvent.press(playlistsCounter);
+
+    await waitFor(() => {
+        expect(screen.getByTestId('library-mpe-rooms-list')).toBeTruthy();
+    });
```
Devessier, line 123: **🎯 suggestion:** ‏This assertion is not sufficient to ensure that the library is displayed when the playlists button is pressed. Indeed, My Library screen is always displayed, as all tabs.

I would suggest to also check that My Profile screen is hidden. This is not the best indication that actually we are displaying My Library screen, but it means that the last screen as been exited.

```suggestion
    await waitFor(() => {
        expect(screen.getByTestId('library-mpe-rooms-list')).toBeTruthy();

        expect(screen.queryByTestId('my-profile-page-container')).toBeNull();
    });
```

Devessier: When I write tests, I try to always verify that tests actually break if the implementation is reverted to its previous state. Often I forget to do so, nonetheless 😆.

Devessier (dismissed):
Nice job!
I suggest one improvement in the test that makes it fail if we revert the implementation.

packages/client/screens/MyProfile/__tests__/MyProfile.test.tsx
prastoin: Yes you're right

**💭 thought:**How could we test the following funnel:
Home -> press button that redirects to library
As the home and library would then both appears to be truthy ?

Devessier: Wow, good question!
It may be possible to use `useIsFocused` hook from React Navigation to conditionally apply a prop to `AppContainer`, and check it in tests with something like `expect(screen).toHaveProp('screen-focused', true)`?

Devessier (approved):
Great! We can merge.

packages/client/screens/MyProfile/__tests__/MyProfile.test.tsx
prastoin: Yep sounds possible :open_mouth:

PR #476, opened 2022-03-01
Replace fake friends in user searchs

Devessier:


packages/client/__tests__/MtvRoomUsersSearch.test.tsx
```diff
@@ -396,7 +393,7 @@ const mtvRoomUsersSearchMachine = mtvRoomUsersSearchModel.createMachine({
                             }: MtvRoomUsersSearchMachineState,
                         ) => {
                             const filteredUsers = filterUsersByNickname(
-                                fakeUsers,
+                                fakeFollowingAndUsers,
```
Devessier, line 396: Please note the distinction with previous code. Following users are part of overall users list. Previously there were two independent lists: friends and users.

packages/client/__tests__/MtvRoomUsersSearch.test.tsx
```diff
@@ -80,31 +80,25 @@ interface TestingContext {
...
-        fill: generateUserSummary,
-    }),
-];
+const fakeUsers = generateArray({
+    minLength: 10,
+    maxLength: 19,
+    fill: () => generateUserSummary({ nickname: `A${internet.userName()}` }),
+});
+
+const fakeFollowingAndUsers = [...fakeFollowing, ...fakeUsers];
```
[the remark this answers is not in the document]
prastoin: Overall I can't find in what way the server handler does have access to this data ?

prastoin (approved):
Well done, not the easiest refactor :smile: 
We can merge :+1:

packages/client/__tests__/MtvRoomUsersSearch.test.tsx
prastoin: I can see that you do test that both fakeFollowing and fakeUsers can occurs on a user search result, but in reality I'm not sure that this would work in the case of a search query matching a following nickname :thinking:

PR #479, opened 2022-03-03
Set adonis auth web and api

prastoin:
In this PR we've set up the adonis auth & session package
We will use both API tokens and web auth authentication method to be able to connect via web browser but also from mobile app
We've updated the already existing User to fit the adonis auth requirements and update the config
We've also created basic test to cover both web auth and api tokens login


Devessier (approved):
This is a really good start about authentication!
We can merge.

Issue #482, opened 2022-03-03
SignUp form and logic


prastoin:
First step ( without email verification ):
client:
- [x] Create a `signUp` screen in the existing `AuthenticationStack`, inside the screen create a form containing an email input username input and password input but also a submit button, parse the email and password on the fly, user should not be able to send a request if email is correctly formated `address@domain.extension` and if the password is not **strong** enough
- [x] On request failure depending on error status ( `username` or `email` ) display specific form error
- [x] On request success the user is authenticated and he should be automatically redirected to the application ~~or ?? user is redirected to the `signIn` screen??~~
- [x] tests

server:
- [x] create a `SignUp` route that takes `userName, email, password, authenticationMode`
- [x] Handle SignUp route, verify `username` and `email` availability to send back custom errors, but also check email is valid and password is **strong**
- [x] After parsing request body, create the user in database, sign in the user depending on the given `authenticationMode` and sendback the authenticated response ~~or ?? send back a 200 response ??~~
- [x] tests

PR #488, opened 2022-03-03
User authentication server signup

prastoin:
related to https://github.com/AdonisEnProvence/MusicRoom/issues/482

TODO:
- [x] migrate e2e tests


packages/server/app/Controllers/Http/AuthenticationController.ts
```diff
@@ -12,22 +15,96 @@ export const SignInRequestBody = z.object({
...
+                    throw new Error(
+                        `unknown authentication mode encountered ${authenticationMode}`,
+                    );
+                }
+            }
+        } catch (e) {
+            console.log(e);
+
+            // if zod fails here we consider it as bad email edge case
+            // as the authenticationMode should not be settable by the user
```
Devessier, line 102: **💭 thought:** ‏I think we can assume an error would be thrown only because of an invalid email, but actually async operations such as `User.findBy` can fail, and ideally I think we should return a 500 response with a generic error status in this case.

We could determine specifically if the email is invalid by using `SignUpRequestBody.safeParse()`, which returns the error without throwing it, which is more convenient for such tests.

**I ask for no change**. These are just some thoughts.

packages/server/tests/AuthenticationSignUp.test.ts
```diff
@@ -15,17 +23,178 @@ test.group('Authentication sign up tests group', (group) => {
...
+        assert.equal(getMeResponseBody.body.user.nickname, userNickname);
+        assert.equal(getMeResponseBody.body.user.email, email);
+        assert.isUndefined(getMeResponseBody.body.user.password);
+    });
+
+    test('It should fail to sign up as given password is weak', async (assert) => {
+        const request = supertest.agent(BASE_URL);
+        const email = internet.email();
+        const userNickname = internet.userName();
+        const password = internet.password();
```
Devessier, line 103: **🎯 suggestion:** ‏I think this is risky to postulate that `internet.password()` returns an invalid password according to our standards, because we absolutely do not control what it returns. I would suggest to create a function such as `generateStrongPassword` that you created, but to return wrong passwords.

packages/server/tests/utils/TestUtils.ts
```diff
@@ -109,6 +109,15 @@ export const TEST_USER_ROUTES_GROUP_PREFIX = '/user';
...
 export const TEST_MY_PROFILE_ROUTES_GROUP_PREFIX = '/me';
 
+const TEST_STRONG_PASSWORD: string[] = [':net66LTW', 'RN4k`d8he9k.'];
+export function generateStrongPassword(): string {
+    return TEST_STRONG_PASSWORD[
+        datatype.number({
+            max: TEST_STRONG_PASSWORD.length - 1,
+        })
+    ];
+}
```
Devessier, line 119: **👏 praise:** ‏Really nice!

packages/types/src/authentication.ts
```diff
@@ -1,7 +1,36 @@
 import * as z from 'zod';
+import { UserSummary } from '.';
 
-export const SignUpResponseBody = z.object({
-    userID: z.string().uuid(),
-    userNickname: z.string(),
+export const AuthenticationModeValues = z.enum(['web', 'api']);
```
Devessier, line 4: **👏 praise:** ‏Nice! I like these denominations.

packages/types/src/authentication.ts
```diff
@@ -1,7 +1,36 @@
...
+export type AuthenticationModeValues = z.infer<typeof AuthenticationModeValues>;
+
+export const SignUpSuccessfullResponseBody = z.object({
+    token: z.string().optional(),
```
Devessier, line 8: **💭 thought:** Maybe will we find interesting, when dealing with front-end code, to have two response types, one for web authentication, with no token, and another one for api authentication, with a mandatory token?

Devessier (approved):
Great job! I want to see the implementation for the client!
I wrote some thoughts and made one suggestion.
We can merge!

packages/types/src/authentication.ts
prastoin: I was also wondering on this point
But pragmatically there's no need to make the distinction if it's not for global comprehension that is 100% a good motivation
I mean that web auth doesnot implies any more code in the frontend, the server will manage it for us. In this way the client can consider that if the token is set then it needs to be stored as it involves an api auth ?
I'm not sure about this way of thinking, but it's still was the one I adopted :laughing:

packages/server/tests/AuthenticationSignUp.test.ts
prastoin: about to be fixed

packages/server/app/Controllers/Http/AuthenticationController.ts
prastoin: Indeed I think this is a great idea
Going deeper

packages/types/src/authentication.ts
Devessier: I think that in the client we will make a clear distinction between web and api authentications, in term of processes to execute depending on the platform.
If we are on the web, we ask for web authentication and expect no token.
If we are on a native platform, we ask for api authentication and expect a token, that we store.
Having strong types would make the contract between frontend and backend even stronger.

prastoin: **💭 thought:** Should we have two distinct routes that could hit the same method by precomputing the authenticationMode in this way we could strongly type specific route responseBody or is it overkill ?
Indeed I wasn't looking it this way, by platform. You're right

prastoin: > thought_balloon thought: Should we have two distinct routes that could hit the same method by precomputing the authenticationMode in this way we could strongly type specific route responseBody or is it overkill ?

I mean It's overkill, typescript and the client signUp service itself will do the job.

Devessier: I think the implementation you made of the server route is great!

prastoin: Updated ! in last commits

packages/server/tests/AuthenticationSignUp.test.ts
```diff
@@ -15,17 +25,192 @@ test.group('Authentication sign up tests group', (group) => {
...
+        assert.equal(getMeResponseBody.body.user.nickname, userNickname);
+        assert.equal(getMeResponseBody.body.user.email, email);
+        assert.isUndefined(getMeResponseBody.body.user.password);
+    });
+
+    test('It should fail to sign up as given password is weak', async (assert) => {
+        const request = supertest.agent(BASE_URL);
+        const email = internet.email();
+        const userNickname = internet.userName();
+        const password = generateWeakPassword();
```
Devessier, line 103: **👏 praise:** ‏👌

packages/server/tests/utils/TestUtils.ts
```diff
@@ -109,6 +109,24 @@ export const TEST_USER_ROUTES_GROUP_PREFIX = '/user';
...
+
+const TEST_WEAK_PASSWORD: string[] = ['bestpasswor@1', 'abcqwerty', 'ABCWE'];
+export function generateWeakPassword(): string {
+    return TEST_WEAK_PASSWORD[
+        datatype.number({
+            max: TEST_WEAK_PASSWORD.length - 1,
+        })
+    ];
+}
```
Devessier, line 128: **👏 praise:** ‏👌

packages/server/app/Controllers/Http/AuthenticationController.ts
```diff
@@ -12,22 +15,106 @@ export const SignInRequestBody = z.object({
 });
 export type SignInRequestBody = z.infer<typeof SignInRequestBody>;
 
+// Inspired by https://stackoverflow.com/questions/5142103/regex-to-validate-password-strength
+// Password has to have 2 caps letter one special char 2 digits and 3 lowercase char and minimal length has to be 8
+// Valid password examples: ':net66LTW', 'RN4k`d8he9k.'
+// Invalid password examples: 'bestpasswor@1', 'abcqwerty', 'ABCWE'
```
Devessier, line 21: **👏 praise:** ‏👌

packages/types/src/authentication.ts
```diff
@@ -1,7 +1,56 @@
...
+            'WEAK_PASSWORD',
+        ]),
+    })
+    .or(SignUpSuccessfullResponseBody);
 export type SignUpResponseBody = z.infer<typeof SignUpResponseBody>;
+
+export const SignUpRequestBody = z.object({
+    userNickname: z.string(),
+    password: z.string(),
+    email: z.string().email({ message: 'INVALID_EMAIL' }).max(255, {
```
Devessier, line 50: **👏 praise:** ‏Perfect!

Devessier (approved):
Perfect!
We can merge.

Devessier (approved):
`localeCompare` is way better!
We can merge.

PR #489, opened 2022-03-07
Refactor sign up server to send failure reason array

prastoin:
Before the signup would failed on the first encountered exception
This is a new need for the UI
The user could have to re submit several time the same form to discover that some fields would fail weirdly between each attempt

### DIFF
--- a/packages/server/tests/AuthenticationSignUp.test.ts
+++ b/packages/server/tests/AuthenticationSignUp.test.ts
@@ -1,11 +1,12 @@
 import Database from '@ioc:Adonis/Lucid/Database';
 import {
     ApiTokensSuccessfullSignUpResponseBody,
+    SignUpFailureResponseBody,
     SignUpRequestBody,
-    SignUpResponseBody,
     WebAuthSuccessfullSignUpResponseBody,
 } from '@musicroom/types';
-import { internet, random } from 'faker';
+import User from 'App/Models/User';
+import { datatype, internet, random } from 'faker';
 import test from 'japa';
 import supertest from 'supertest';
 import urlcat from 'urlcat';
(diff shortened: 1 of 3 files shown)


packages/server/tests/AuthenticationSignUp.test.ts
```diff
@@ -230,7 +253,55 @@ test.group('Authentication sign up tests group', (group) => {
...
+        [
+            'UNAVAILABLE_NICKNAME',
+            'UNAVAILABLE_EMAIL',
+            'INVALID_EMAIL',
+            'WEAK_PASSWORD',
+        ].forEach((reasonValue) =>
+            assert.isDefined(
+                signFailureReasons.find((error) => error === reasonValue),
+            ),
+        );
```
Devessier, line 305: **
**🎯 suggestion:** This is a matter of coding style, but this could be written in a more concise way, like so:

```suggestion
        assert.deepEqual(signFailureReasons, ['UNAVAILABLE_NICKNAME', 'UNAVAILABLE_EMAIL', 'INVALID_EMAIL', 'WEAK_PASSWORD' ]);
```

packages/types/src/authentication.ts
```diff
@@ -32,24 +32,34 @@ export type SignUpSuccessfullResponseBody = z.infer<
...
+export type SignUpFailureReasons = z.infer<typeof SignUpFailureReasons>;
+
+export const SignUpFailureResponseBody = z.object({
+    status: z.literal('FAILURE'),
+    signFailureReasons: SignUpFailureReasons.array(),
+});
+
+export type SignUpFailureResponseBody = z.infer<
+    typeof SignUpFailureResponseBody
+>;
```
Devessier, line 51: **👏 praise:** ‏Good job, this is great to handle errors that way! Back-end code done with quality in mind is interesting to read.

Devessier (dismissed):
Nice job!
We are entering in the world of back-end code that returns meaningful and semantic HTTP errors. Love it!
I made suggestions about some test assertions, feel free to discard them.
We can merge.

Devessier (approved):
Nice!
We can merge.

Issue #491, opened 2022-03-08
Finalize global authentication and signing in


Devessier:
## Todo

- [x] Type `/authentication/sign-in` request body
- [x] Type `/authentication/sign-in` response body
- [x] Save token returned by `/sign-in` to AsyncStorage if applicable
- [x] Send authorization token alongside every request, if applicable (on native platforms)
- [x] Migrate from `/authentication/me` to `/me/profile-information`
- [x] Handle errors in `/authentication/sign-in` as @prastoin did for `/authentication/sign-up`
- [x] Delete `/authentication/me` route and corresponding Controller method

PR #493, opened 2022-03-09
Refine global authentication and signing in

Devessier:
Closes https://github.com/AdonisEnProvence/MusicRoom/issues/491

### DIFF
--- a/packages/client/App.tsx
+++ b/packages/client/App.tsx
@@ -3,7 +3,7 @@ import { BottomSheetModalProvider } from '@gorhom/bottom-sheet';
 import { DripsyProvider, Text, View } from 'dripsy';
 import { StatusBar } from 'expo-status-bar';
 import React, { useState } from 'react';
-import { ImageBackground } from 'react-native';
+import { ImageBackground, Platform } from 'react-native';
 import { SafeAreaProvider } from 'react-native-safe-area-context';
 import { QueryClient, QueryClientProvider } from 'react-query';
 import { AppContextProvider } from './contexts/AppContext';
@@ -84,7 +84,10 @@ const App: React.FC = () => {
                         </BottomSheetModalProvider>
                     </SafeAreaProvider>
                 </DripsyProvider>
-                <ReactQueryDevtools initialIsOpen={false} />
+
+                {Platform.OS === 'web' ? (
+                    <ReactQueryDevtools initialIsOpen={false} />
+                ) : null}
             </QueryClientProvider>
         );
     }
--- a/packages/client/machines/appMachine.ts
+++ b/packages/client/machines/appMachine.ts
@@ -8,8 +8,9 @@ import {
     StateMachine,
 } from 'xstate';
 import { SocketClient } from '../contexts/SocketContext';
-import { getMe, sendSignIn } from '../services/AuthenticationService';
+import { sendSignIn } from '../services/AuthenticationService';
 import { request } from '../services/http';
+import { getMyProfileInformation } from '../services/UsersSearchService';
 import { appModel } from './appModel';
 import { createAppMusicPlayerMachine } from './appMusicPlayerMachine';
 import { createAppMusicPlaylistsMachine } from './appMusicPlaylistsMachine';
@@ -233,7 +234,7 @@ export function createAppMachine({
         {
             services: {
                 fetchUser: async () => {
-                    const me = await getMe();
+                    const me = await getMyProfileInformation();
 
                     return me;
                 },
--- a/packages/client/machines/appMachine.ts
+++ b/packages/client/machines/appMachine.ts
@@ -1,3 +1,4 @@
+import { Platform } from 'react-native';
 import invariant from 'tiny-invariant';
 import {
     ContextFrom,
@@ -8,6 +9,7 @@ import {
 } from 'xstate';
 import { SocketClient } from '../contexts/SocketContext';
 import { getMe, sendSignIn } from '../services/AuthenticationService';
+import { request } from '../services/http';
 import { appModel } from './appModel';
 import { createAppMusicPlayerMachine } from './appMusicPlayerMachine';
 import { createAppMusicPlaylistsMachine } from './appMusicPlaylistsMachine';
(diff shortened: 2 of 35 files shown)


packages/server/app/Middleware/EveryAuth.ts
```diff
@@ -0,0 +1,11 @@
+import { HttpContextContract } from '@ioc:Adonis/Core/HttpContext';
+import AuthMiddleware from './Auth';
+
+export default class EveryAuthMiddleware extends AuthMiddleware {
+    public async handle(
+        context: HttpContextContract,
+        next: () => Promise<void>,
+    ): Promise<void> {
+        return super.handle(context, next, ['web', 'api']);
```
prastoin, line 9: praise: so powerfull

packages/server/app/Controllers/Http/MyProfileController.ts
```diff
@@ -1,21 +1,17 @@
...
 export default class MyProfileController {
     public async getMyProfileInformation({
-        request,
+        auth,
     }: HttpContextContract): Promise<GetMyProfileInformationResponseBody> {
-        const rawBody = request.body();
+        const user = auth.user;
+        invariant(
+            user !== undefined,
+            'User must be authenticated to get her profile information',
```
prastoin, line 12: **🤓 nitpick:** I'm not sure if the error message is accurate. Unless I'm mistaken if the user is not authenticated then adonis would throw an error before ?
I have no clue on how should be this error but at the first read it looks like the authentication assertion is made manually here and not by the adonis auth middleware. Let me know what you think

packages/server/app/Controllers/Http/MyProfileController.test.ts
```diff
@@ -34,27 +60,24 @@ test.group('Users Profile information tests', (group) => {
...
     });
 
-    test('It should retrieve my profile information', async (assert) => {
-        const userID = datatype.uuid();
-        const userNickname = internet.userName();
-        await createUserAndGetSocket({
-            userNickname,
-            userID,
+    test('Retrieves profile information of the current authenticated user', async (assert) => {
+        const request = supertest.agent(BASE_URL);
```
prastoin, line 64: **👏 praise:** really nice, it also works if not iterating on already existing seeded user

packages/server/app/Controllers/Http/MyProfileController.test.ts
```diff
@@ -63,47 +86,37 @@ test.group('Users Profile information tests', (group) => {
...
-            userNickname,
+            userID: user.uuid,
+            userNickname: user.nickname,
         };
         assert.deepEqual(parsedBody, expectedBody);
     });
 
-    test('It should send back user not found', async () => {
-        const userID = datatype.uuid();
+    test('Returns a 401 error when the current user is unauthenticated', async () => {
```
prastoin, line 95: 💭 thought: interesting on how adding the authentication changes our tests definition

packages/client/App.tsx
```diff
@@ -84,7 +84,10 @@ const App: React.FC = () => {
                         </BottomSheetModalProvider>
                     </SafeAreaProvider>
                 </DripsyProvider>
-                <ReactQueryDevtools initialIsOpen={false} />
+
+                {Platform.OS === 'web' ? (
```
prastoin, line 88: ❔ question: Maybe should we also check for dev env ?

prastoin: In fact it won't last long anw

packages/client/tests/server/handlers.ts
```diff
@@ -40,11 +41,44 @@ import {
...
+    RestContext,
+    RestRequest,
+} from 'msw';
 import * as z from 'zod';
 import { SERVER_ENDPOINT } from '../../constants/Endpoints';
 import { SearchTracksAPIRawResponse } from '../../services/search-tracks';
 import { db } from '../data';
+import { testGetFakeUserID } from '../tests-utils';
+
+function withAuthentication<
```
prastoin, line 58: 👏 praise: awesome

packages/client/services/http.ts
```diff
@@ -0,0 +1,92 @@
...
+    clearToken(): Promise<void>;
+};
+
+/**
+ * Creates a custom instance of redaxios that implements methods
+ * to handle authentication by opaque tokens.
+ *
+ * This function must not be used outside of tests.
+ */
+export function createRequester(): Requester {
```
prastoin, line 34: 👏 praise: I'm falling in love

packages/client/services/AuthenticationService.ts
```diff
@@ -1,47 +1,69 @@
...
+        password,
+        authenticationMode: 'api',
+    };
+
+    const response = await request.post('/authentication/sign-in', body);
+    const parsedResponse = SignInSuccessfulApiTokensResponseBody.parse(
+        response.data,
     );
+
+    await request.persistToken(parsedResponse.token);
```
prastoin, line 64: 👏 praise: hehe

packages/client/machines/appMachine.ts
```diff
@@ -45,9 +48,25 @@ export function createAppMachine({
         {
             id: 'app',
 
-            initial: 'fetchingInitialUserAuthenticationState',
+            initial: 'loadingAuthenticationTokenFromAsyncStorage',
 
             states: {
+                loadingAuthenticationTokenFromAsyncStorage: {
```
prastoin, line 54: *Async*Storage is aptly named

packages/client/machines/appMachine.ts
```diff
@@ -45,9 +48,25 @@ export function createAppMachine({
...
+                    always: {
```
prastoin, line 55: ❔ question: Should this state also have the tags `showApplicationLoader` ?

prastoin (dismissed):
Wow so much done on this one !
Really nice work
Leaved several comments
Haven't seen any major can already approve this PR

packages/server/app/Controllers/Http/MyProfileController.ts
Devessier: I think that the method of the controller has no idea whether an authentication middleware has been called before it.
At least TypeScript does not, and `auth.user` can be `undefined`.
This is why I named the error that way, because there is no strict link contract between the method and the route, which requires at its level the auth middleware.

prastoin (approved):
Mergin when ci is good !

Issue #498, opened 2022-03-10
Handle signing in errors more precisely


Devessier:
Two types of errors can be received:

- Credentials invalid
- An unknown error occured (probably with a 500 status)

We want to dissociate the text displayed when an error occurs.

PR #499, opened 2022-03-11
Differentiate invalid credentials error from unknown errors

Devessier:
Closes https://github.com/AdonisEnProvence/MusicRoom/issues/498

### DIFF
--- a/packages/client/__tests__/Authentication.test.tsx
+++ b/packages/client/__tests__/Authentication.test.tsx
@@ -34,7 +34,7 @@ interface TestingContext {
 const existingUser = generateAuthenticationUser();
 
 const authenticationModelMachine =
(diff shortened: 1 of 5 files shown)


packages/client/__tests__/Authentication.test.tsx
```diff
@@ -458,31 +458,71 @@ const authenticationModelMachine =
...
+                                                }: TestingContext) => {
+                                                    invariant(
+                                                        screen !== undefined,
+                                                        'Screen must have been rendered',
+                                                    );
+
+                                                    await waitFor(() => {
+                                                        const serverErrorAlert =
+                                                            within(
+                                                                screen.getByTestId(
```
prastoin, line 508: Not important at all but on the sign up screen we do display a toast error, maybe we should pick at least one common solution for both screens ?
Might be later

prastoin (approved):
Great shaping !
Ready to be merged

packages/client/__tests__/Authentication.test.tsx
Devessier: Indeed, we will have to normalize the UX of the authentication screens.
I did not want to take these decisions alone so I went to the simplest solution, according to the code I already had.

prastoin: We will then determine all of this details later during our UI/UX improvements journey

PR #500, opened 2022-03-11
Use real authentication for /search/tracks and /search/rooms

Devessier:


packages/server/tests/utils/TestUtils.ts
```diff
@@ -592,6 +600,54 @@ export function initTestUtils(): TestUtilsReturnedValue {
         }
     }
 
+    async function createUserAndAuthenticate(
```
prastoin, line 603: Praise: great !

packages/client/machines/searchMtvRoomsMachine.ts
```diff
@@ -151,13 +151,9 @@ export const searchMtvRoomsMachine = searchMtvRoomsModel.createMachine(
                 ({ nextPage, searchQuery }) =>
                 async (sendBack) => {
                     try {
```
prastoin, line 153: Wow new area begins !

PR #501, opened 2022-03-11
Use real auth remaining ungrouped routes

Devessier:


packages/types/src/mpe-room-search.ts
```diff
@@ -22,7 +21,6 @@ export type MpeSearchMyRoomsResponseBody = z.infer<
 >;
 
 export const ListAllMpeRoomsRequestBody = z.object({
-    userID: z.string().uuid(), // FIXME: use authentication
```
prastoin, line 25: praise: satisfaction !

PR #509, opened 2022-03-16
Create E2E test for signing up

Devessier:
Closes https://github.com/AdonisEnProvence/MusicRoom/issues/507


prastoin (approved):
Looks good to me !
Could maybe test that the user is not authenticated when creating a whole new page ?
Anw can already approve

Issue #483, opened 2022-03-03
SignOut button and logic


prastoin:
client: 
- [x] add a `signOut` button inside the `MyProfileSettings` screen, on press it will send a `signOut` request
- [x] add a `signOut` button inside the `userEmailVerification` screen, on press it will send a `signOut` request
- [x] On a `signOut` response remove api auth token from `tokenAsyncStorage`, Also the application itself should redirects the user to the authentication initial screen
- [x] e2e test for signout from my profile settings screen
- [x] e2e test for sign out from email verification screen
- [x] integ tests for sign out from my profile settings screen
- [x] integ tests for sign out from email verification screen

server:
- [x] create a `signOut` guarded route
- [x] Handle the `signOut` route that will use the `auth.logout` web auth method to remove session cookie, if a api token is provided it will use `auth.revoke(token)`
- [x] tests

PR #512, opened 2022-03-17
Sign out implem client

prastoin:
related to #483


Devessier (approved):
Nice job!
We can merge!

Issue #514, opened 2022-03-17
Use broadcast channel to spread signIn signOut and signUp to every browser tabs


prastoin:
Client:
- [x] use broadcastChannel to persist to every browser same origin tab each signOut signIn signUp through an event called RELOAD_BROWSER_TAB
- [x]  inside appMachine listen to the broadcastChannel event RELOAD_BROWSER_TAB on receive reload the tab
- [x] e2e test sign out multi tabs
- [x] e2e test sign in multi tabs
- [x] e2e test sign up multi tabs

PR #515, opened 2022-03-17
Implem broadcast channel on web to perform sign up/in/out on every existing tabs

prastoin:
related to #514


packages/client/contexts/AppContext.tsx
```diff
@@ -109,6 +110,9 @@ export const AppContextProvider: React.FC<MusicPlayerContextProviderProps> = ({
         () =>
             createAppMachine({
                 socket,
+                musicRoomBroadcastChannel: PLATFORM_OS_IS_WEB
+                    ? new BroadcastChannel('music_room_channel')
```
Devessier, line 114: **👏 praise:** ‏Nice! I've never used `BroadcastChannel` 😮

Devessier (approved):
This is an amazing PR!
It's so nice to see a `BroadcastChannel` in use! Love that.
We can merge.

PR #521, opened 2022-03-21
Dump react-native-toast & fix mpe export as mtv with constraints

prastoin:
close https://github.com/AdonisEnProvence/MusicRoom/issues/520
also handling:
1/ react-native-toast-message deprecated requirements
2/ Fail to display image


prastoin:
From there we're having a blocking issue.
The temporal test looks like he's failing as he's comparing pointer addresses inside the activity mock and this only in the CI..
We've already spent a lot of time on this, this is not major in this way we have to find a new way to test the same assertion that is: user can create an mtv room with constraint from an mpe export

We will leave in the temporal integ test a simple test that will attest that perform an export with specific constraint will lead to be calling the expected activity but we won't check it's args
To be able to test that the room is exported with constraints we have to establish an e2e test

Devessier (approved):
Perfect!
We can merge.

Issue #486, opened 2022-03-03
PLAN - email verification


prastoin:

prastoin:
We've been talking about e2e testing and email verification
We were thinking about using an online temp mail website to be able to receive an scrap the email verification token
In the best of the world we would to be dependent from this external website on only one test an not the entire e2e test coverage.
In this way we should also determine a e2e custom way to allow the client to either retrieve the latest token or by pass the email verification

prastoin:
We should not forget about testing the apparition of the email verification test concurrency
ex 2 devices both on email verification one verifies what happens on the second
User that has signed up but didn't verify his email, sign in later on an other device we should still display the email verification screen

prastoin:
> We should not forget about testing the apparition of the email verification test concurrency ex 2 devices both on email verification one verifies what happens on the second User that has signed up but didn't verify his email, sign in later on an other device we should still display the email verification screen

When a user has verified his email, every screen rendering the email verification form should poll the information to the server as the email verification is a global account information ( just to say no device related like could be a sign in/up/out )

Devessier:
All tasks are done!

PR #524, opened 2022-03-23
Sign up server send confirmation email

prastoin:
related to https://github.com/AdonisEnProvence/MusicRoom/issues/486


packages/server/tests/AuthenticationSignUp.test.ts
```diff
@@ -24,11 +27,14 @@ test.group('Authentication sign up tests group', (group) => {
         initTestUtils();
 
     group.beforeEach(async () => {
+        //Init mail.trap here to prevent sending mails to random email if .env is correctly set see test `It should fail to sign up as user with given email is already in base`
+        Mail.trap(noop);
```
Devessier, line 31: **👏 praise:** ‏👌

packages/server/tests/AuthenticationSignUp.test.ts
```diff
@@ -70,12 +93,86 @@ test.group('Authentication sign up tests group', (group) => {
...
+                message.subject,
+                `Welcome ${nickname}, please verify your email !`,
+            );
+        });
+
+        const emailVerification = new EmailVerification(user, 'token1234');
+        await emailVerification.send();
+
+        //Mail.trap will be expected without any more required actions such as waitFor
+    });
```
Devessier, line 128: See my comment above about email sending assertion.

packages/server/resources/views/emails/email_verification.edge
```diff
@@ -0,0 +1,21 @@
+<mjml>
```
Devessier, line 1: **❔ question:** ‏What resources did you use to learn MJML?

packages/server/resources/views/emails/email_verification.edge
```diff
@@ -0,0 +1,21 @@
...
+  <mj-body>
+    <mj-section>
+      <mj-column>
+
+        <mj-text align="left" color="#F45E43" font-family="Arial, sans-serif" font-size="13px" line-height="22px" padding-bottom="5px" padding-top="25px" padding="10px 25px">
+          <p style="line-height: 60px; text-align: center; margin: 10px 0;font-size:55px;color:#212121;font-family:'Times New Roman',Helvetica,Arial,sans-serif"><b>Musicroom</b></p>
```
Devessier, line 7: **❔ question:** ‏I do not know MJML. Why do you use a `p` tag to display text here, with inline styles, while below you directly put text in a `mj-text` component?

packages/server/tests/AuthenticationSignUp.test.ts
```diff
@@ -39,6 +45,23 @@ test.group('Authentication sign up tests group', (group) => {
...
+
+            assert.deepEqual(message.from, {
+                address: 'no-reply@adonisenprovence.com',
+            });
+
+            assert.equal(
+                message.subject,
+                `Welcome ${userNickname}, please verify your email !`,
+            );
+        });
```
Devessier, line 63: **❔ question:** ‏I commented the line that sends the email in `signUp` method, and tests did not break. We should probably use `assert.plan()` to ensure that assertions are made?

But I do not understand how the test can work without a kind of `waitFor`, because Japa can not know when we send the email we want to assert on. We could send the email after we returned the HTTP response for instance.

packages/server/tests/AuthenticationSignUp.test.ts
```diff
@@ -70,12 +93,86 @@ test.group('Authentication sign up tests group', (group) => {
...
+
+            Mail.trap(() => {
+                const onSignUpFailEmailShouldNotBeSent = true;
+                assert.isFalse(onSignUpFailEmailShouldNotBeSent);
+            });
+
+            const emailVerification = new EmailVerification(user, 'token1234');
+            await emailVerification.send();
+        },
+    );
```
Devessier, line 151: **❔ question:** ‏Why should this test fail? What does make it different from the very above test?

packages/server/tests/AuthenticationSignUp.test.ts
```diff
@@ -70,12 +93,86 @@ test.group('Authentication sign up tests group', (group) => {
...
+
+            assert.deepEqual(message.from, {
+                address: 'no-reply@adonisenprovence.com',
+            });
+
+            assert.equal(
+                message.subject,
+                `Welcome ${userNickname}, please verify your email !`,
+            );
+        });
```
Devessier, line 174: See my comment above about email sending assertion.

packages/server/tests/AuthenticationSignUp.test.ts
```diff
@@ -309,6 +439,23 @@ test.group('Authentication sign up tests group', (group) => {
...
```
Devessier, line 457: See my comment above about email sending assertion.

Devessier (commented):
Good job! Nice to see how Adonis supports emails.
I asked some questions to better understand how tests work.

packages/server/tests/AuthenticationSignUp.test.ts
prastoin: Here I'm testing that the following assertion:
```ts
        Mail.trap(() => {
                const expectingNoEmailToBeReceived = true;
                assert.isFalse(expectingNoEmailToBeReceived);
            });
```
Is failing when an email is received inside the Mail.trap. In this way I can be sure that when I'm using it inside test that should not receive email it would fail after receiving one

packages/server/tests/AuthenticationSignUp.test.ts
prastoin: Interesting ! digging deeper

prastoin: I achieved to be using sinon.spy to verify that the spy is called on the test end

packages/server/resources/views/emails/email_verification.edge
prastoin: I paid someone on fiver !
Seriously I've just look at existing templates https://mjml.io/templates

packages/server/tests/AuthenticationSignUp.test.ts
prastoin: Anyway ended up refactoring everything using sinon.spy

packages/server/resources/views/emails/email_verification.edge
prastoin: I've been using a template centered text section for the token and title

Devessier (approved):
Nice job!
We can merge.

Issue #525, opened 2022-03-23
e2e email verification retrieve token


prastoin:
See https://github.com/AdonisEnProvence/MusicRoom/issues/486 comments to understand this issue's redaction

We need to be able to retrieve a dynamic token in the e2e tests
There could be two ways to do so:

## Alive inbox
We could open a temp email page retrieve the generated email and sign up using it, then on sign up the server will send a email confirmation mail to the given email ( that is alive in a of the browser tab ).
We would scrap the open page and retrieve the token from the received email
This would be quite wobbly because we would be dependent on on an external website such as [temp-mail](https://temp-mail.org/fr/) or [mail7](https://mail7.io/)
This could still be a good solution associated with the second solution the `e2e test private route`

## e2e test private route
We could set up a route inside the server api that would return the last created email verification token for a given email.
This route should be guarded, this would allow to retrieve related authenticated user email confirmation token easily.
This route should only be available in e2e tests, see .env variables ?
This route would be called via playwright requester, which means inside e2e test context only

### EDIT
We've completly forgotten about the fact that we will be storing hashed token in databse, in this way it's impossible to retrieve it after it's creation.
To face this issue, we will still implement an authenticated e2e test specific route that will change the authenticated user isEmailConfirmed value directly in database without awaiting for any token

## Conclusion
In my opinion better the coverage will be if we achieve to be using both of the solutions !
Only one e2e test, that could be a specific sign-up e2e test would be using the described `Alive inbox` and every other one would be using the `e2e test private route` via centralized testUtils methods

### Sending email in e2e tests inside testUtils performSignup
We've discovered that while running any e2e test, as they all perform a sign up an email will be sent to a random generated email. This is not cool.
To face this issue, we need to create a route to be able to toggle Mail.trap() ( it this works ).
The route would be called something like `/test/toggle-mail-trap`

Devessier:
I agree with you! We can use both solutions.

prastoin:
Closing issues as it has been resolved

PR #528, opened 2022-03-24
Verify email e2e test

prastoin:
Done here
- Alive inbox
- e2e test private route
see https://github.com/AdonisEnProvence/MusicRoom/issues/525

### DIFF
--- a/packages/client/package.json
+++ b/packages/client/package.json
@@ -21,7 +21,9 @@
     "test": "jest --reporters jest-silent-reporter",
     "test:watch": "jest --watch",
     "install:react-youtube": "cd ../../node_modules/react-youtube && yarn && yarn compile",
-    "test:e2e": "playwright test"
+    "test:e2e": "playwright test",
+    "clean": "rimraf dist *.tsbuildinfo"
+
   },
   "jest": {
     "preset": "jest-expo"
--- a/packages/server/app/Controllers/Http/AuthenticationController.ts
+++ b/packages/server/app/Controllers/Http/AuthenticationController.ts
@@ -68,11 +68,8 @@ export default class AuthenticationController {
             password,
         });
 
-        const emailVerification = new EmailVerification(
-            createdUser,
-            'token1234',
-        );
-        await emailVerification.sendLater();
+        const emailVerification = new EmailVerification(createdUser, '123456');
+        await emailVerification.sendLater().catch((e) => console.error(e));
 
         const { nickname, uuid: userID } = createdUser;
         const userSummary: UserSummary = {
--- a/packages/server/tests/AuthenticationSignUp.test.ts
+++ b/packages/server/tests/AuthenticationSignUp.test.ts
@@ -13,6 +13,7 @@ import { datatype, internet, random } from 'faker';
 import test from 'japa';
 import Sinon from 'sinon';
 import supertest from 'supertest';
+import invariant from 'tiny-invariant';
 import urlcat from 'urlcat';
 import {
     BASE_URL,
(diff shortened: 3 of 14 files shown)


prastoin:
server integ tests are failing due to ci `NODE_ENV` value.
We should either modify the test to run only when the `NODE_ENV` is in `development` or change ci `NODE_ENV` value

packages/server/tests/AuthenticationSignUp.test.ts
```diff
@@ -62,10 +63,15 @@ test.group('Authentication sign up tests group', (group) => {
                 address: 'no-reply@adonisenprovence.com',
             });
 
-            assert.equal(
-                message.subject,
-                `Welcome ${userNickname}, please verify your email !`,
+            const emailVerificationObjectRegex = new RegExp(
+                `[.*].*Welcome.*${userNickname},.*please.*verify.*your.*email.*!`,
```
Devessier, line 67: **💭 thought:** ‏I think the square brackets at the beginning of the Regex do not behave the way you want. See https://regex101.com/r/cwVrtv/1.

The square brackets contain a list of characters that match. I think you should escape them, like that: https://regex101.com/r/CdgogL/1.

packages/server/tests/AuthenticationSignUp.test.ts
```diff
@@ -62,10 +63,15 @@ test.group('Authentication sign up tests group', (group) => {
...
             );
+            assert.isDefined(message.subject);
+            invariant(
+                message.subject !== undefined,
+                'message subject is undefined',
+            );
+            assert.isTrue(emailVerificationObjectRegex.test(message.subject));
```
Devessier, line 74: **🎯 suggestion:** You can use `assert.match()` to test a string against a Regex. It does the same thing as here. 😄

Devessier (approved):
Really nice job!
I love the way you tested that an email with a confirmation code is really received!
I have some questions, but we can merge!

packages/client/package.json
```diff
@@ -21,7 +21,9 @@
     "test": "jest --reporters jest-silent-reporter",
     "test:watch": "jest --watch",
     "install:react-youtube": "cd ../../node_modules/react-youtube && yarn && yarn compile",
-    "test:e2e": "playwright test"
+    "test:e2e": "playwright test",
+    "clean": "rimraf dist *.tsbuildinfo"
```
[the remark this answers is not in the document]
prastoin: I've been having an issue with playwright, this one to be exact https://github.com/microsoft/playwright/issues/7678
As described in the issue thread the problem could come from the build, in my case it wasn't but I still left this script there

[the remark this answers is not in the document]
prastoin: While debugging I needed to clean everything up, it was missing

packages/server/app/Controllers/Http/AuthenticationController.ts
```diff
@@ -68,11 +68,8 @@ export default class AuthenticationController {
             password,
         });
 
-        const emailVerification = new EmailVerification(
-            createdUser,
-            'token1234',
-        );
-        await emailVerification.sendLater();
+        const emailVerification = new EmailVerification(createdUser, '123456');
+        await emailVerification.sendLater().catch((e) => console.error(e));
```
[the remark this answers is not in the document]
prastoin: I think that yes ! the user will  still be able to resend an email after all. I don't really how sendLater would behave if it process and fails just after being put in queue

packages/server/tests/AuthenticationSignUp.test.ts
prastoin: I've been doing so and eslint was triggering `useless escaped chars` errors so I've been removing them
Should we:
- start the regex using .* ?
- escape them and update eslint ?

packages/server/tests/AuthenticationSignUp.test.ts
prastoin: About to be fixed

packages/server/tests/AuthenticationSignUp.test.ts
Devessier: Do you remember what was the Regex exactly? Because escaping characters here is absolutely not useless. 😮

Devessier (approved):
Nice job! It's amazing to discord nice features of Playwright such as `page.request`!
We can merge.

PR #532, opened 2022-03-24
Email verification transaction status polling

prastoin:


packages/client/e2e/_utils/page.ts
```diff
@@ -216,10 +216,6 @@ export async function bypassVerifyEmailScreen({
 
     await page.request.get(`${SERVER_ENDPOINT}/test/bypass-email-confirmation`);
 
-    //FIXME TMP should be removed after email polling
-    await page.reload();
```
Devessier, line 220: **👏 praise:** ‏It seems to be a little deletion, but it is due to a massive improvement.

packages/client/machines/appMachine.ts
```diff
@@ -291,119 +291,198 @@ export function createAppMachine({
...
+                                        onError: {
+                                            target: 'debouncing',
                                         },
-                                    }),
+                                    },
+                                },
+
+                                debouncing: {
+                                    after: {
+                                        500: {
```
Devessier, line 438: **🎯 suggestion:** ‏I think 500ms is waaaaay too short. 5s would be better. I think it's better to let users wait some seconds than making two requests per second in background, for an undetermined amount of time. The application could be made slow for users with a bad internet connection.

packages/client/machines/appMachine.ts
```diff
@@ -291,119 +291,198 @@ export function createAppMachine({
...
+                                    on: {
+                                        SIGN_OUT: undefined,
+                                    },
                                 },
                             },
 
                             on: {
-                                SIGN_OUT: undefined,
+                                SIGN_OUT: {
+                                    target: '.performSignOut',
```
Devessier, line 481: **👏 praise:** ‏Using an internal transition is definitely a good idea! I makes `SIGN_OUT: undefined` useless.

Devessier (dismissed):
Good job! I love your polling implementation.
I think we should make delay between two polling requests longer before merging.

packages/client/__tests__/Authentication.test.tsx
```diff
@@ -1875,6 +1883,38 @@ const authenticationModel = createModel<TestingContext>(
...
+                            playlistsCounter: user.playlistsCounter,
+                            followersCounter: user.followersCounter,
+                            followingCounter: user.followingCounter,
+                            devicesCounter: user.devicesCounter,
+                            hasConfirmedEmail: true,
+                        }),
+                    );
+                }),
+            ),
+        );
```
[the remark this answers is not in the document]
prastoin: Ahhh yes indeed ! I was wondering on how I could be able to retrieve the user we're impersonating !
Indeed gonna make the modication !

Devessier (approved):
Nice job!
We can merge.

PR #537, opened 2022-03-25
Request confirmation email resending

Devessier:


packages/server/app/Controllers/Http/AuthenticationController.ts
```diff
@@ -247,6 +248,8 @@ export default class AuthenticationController {
             'resendConfirmationEmail',
         );
         if (hasReachedRateLimit === true) {
+            response.status(429);
```
prastoin, line 251: Http status code, just like english irregular verbs should be learned at school !

packages/client/machines/appMachine.ts
```diff
@@ -453,6 +456,50 @@ export function createAppMachine({
                             },
                         },
 
+                        resendingConfirmationEmail: {
```
prastoin, line 459: Thoughts: This means that even if the user has confirmed her email or user is not on email confirmation screen he could ask for email resend ?

packages/client/__tests__/AuthenticationResendingConfirmationEmail.test.tsx
```diff
@@ -0,0 +1,258 @@
...
+import { withAuthentication } from '../tests/server/handlers';
+import { db, generateAuthenticationUser } from '../tests/data';
+
+interface TestingContext {
+    screen: ReturnType<typeof render>;
+}
+
+const existingUser = generateAuthenticationUser();
+
+/**
```
prastoin, line 25: Praise: great !

prastoin:
Great implementation ! waiting for CI to be fixed before approving and merging !

packages/client/machines/appMachine.ts
Devessier: `resendingConfirmationEmail` state is at the same level  as `formHandler` and `pollingHandler` states inside `waitingForUserEmailConfirmation` state. I think what you describe can not happen 🤔 Do I miss something?

packages/server/app/Controllers/Http/AuthenticationController.ts
Devessier: Lol, actually Github Copilot suggested me the status code. 😆

packages/client/machines/appMachine.ts
prastoin: Ah my bad !

prastoin (approved):
Ready to be merged !

PR #543, opened 2022-03-28
Add info about tacit acceptance of terms of service

Devessier:


prastoin (approved):
There's not link for the moment to any PP I guess we will be adding that in the future :smile: 
Ready to be merged !

Issue #494, opened 2022-03-09
Trim text filled by the user that is going to be displayed as is


prastoin:
We want to ensure inputs are trimmed. We want to perform this operation only on the backend.

- Trim at the beginning and at the end
- Require at least one character (after trimming)

## Todo

- [ ] Apply to username
- [ ] Apply to email
- [ ] Apply to MTV room name
- [ ] Apply to MPE room name

prastoin:
also see about trimming given userNickname etc etc

PR #547, opened 2022-03-29
Parse mtv mpe rooms names user email and nickname server

prastoin:
close https://github.com/AdonisEnProvence/MusicRoom/issues/494


Devessier (approved):
Nice job!
We can merge.
Front-end tests take toooo much time to run, I think we will have to speed them up 😮

PR #549, opened 2022-03-31
Create route start password reset procedure

Devessier:
What is in this PR:

- New route to start a password reset procedure
- Refactoring of all tests that use `Mail.trap()` where assertions may fail without tests breaking

### DIFF
--- a/packages/server/resources/views/emails/layouts/main.edge
+++ b/packages/server/resources/views/emails/layouts/main.edge
@@ -1,0 +1,15 @@
+<mjml>
+    <mj-body background-color="white">
+        <mj-section>
+            <mj-column>
+                <mj-text align="center" font-family="helvetica" font-size="40px" padding-top="25px">
+                    Musicroom
+                </mj-text>
+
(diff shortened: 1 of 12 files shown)


packages/server/start/routes.ts
```diff
@@ -234,6 +238,19 @@ if (nodeEnvIsDevelopment) {
             'TestEnvMethodController.toggleMailTrap',
         );
     }).prefix(TEST_GROUP_PREFIX);
+
+    Route.get('/email/:template', async ({ request, view }) => {
```
prastoin, line 242: prase: interesting !

packages/server/resources/views/emails/layouts/main.edge
```diff
@@ -0,0 +1,15 @@
+<mjml>
```
prastoin, line 1: praise: great idea

prastoin (approved):
Really nice layouts for the amails mjml and interesting way to display email in local
Ready to be merged

PR #550, opened 2022-04-01
Add button to request password reset

Devessier:


packages/client/services/AuthenticationService.ts
```diff
@@ -214,3 +215,28 @@ export async function sendResendingConfirmationEmail(): Promise<ResendConfirmati
...
+}: SendRequestingPasswordResetArgs): Promise<RequestPasswordResetResponseBody> {
+    const requestBody: RequestPasswordResetRequestBody = {
+        email,
+    };
+    const rawResponse = await request.post(
+        '/authentication/request-password-reset',
+        requestBody,
+        {
+            validateStatus: (status) =>
+                status === 200 || status === 404 || status === 429,
```
prastoin, line 234: praise: I think we should always be doing something like that !

prastoin: We never knows what exception adonis could throw between 200 and 429

packages/client/screens/SigningInScreen.tsx
```diff
@@ -223,6 +240,17 @@ const SigningInScreen: React.FC<SigningInScreenProps> = ({ navigation }) => {
...
                                 />
 
+                                <TouchableOpacity
+                                    style={sx({
+                                        marginTop: 'm',
+                                    })}
+                                    onPress={handleRequestPasswordReset}
+                                >
+                                    <Text sx={{ color: 'white' }}>
+                                        Do you lost your password?
```
prastoin, line 250: nitpick: Unless I'm mistaken should this be `Did you lost your password ?` or maybe we could be saying something like `Reset password`

Devessier: I made a typo you're right, but I think it should be _Did you lose your password_? Or _Have you lost your password_?

prastoin: Indeed :laughing:

PR #551, opened 2022-04-01
Create route to validate a password reset token

Devessier:


packages/server/app/Controllers/Http/AuthenticationController.ts
```diff
@@ -315,4 +317,39 @@ export default class AuthenticationController {
...
+        });
+        const isInvalidToken = isValidToken === false;
+        if (isInvalidToken === true) {
+            response.status(400);
+
+            return {
+                status: 'INVALID_TOKEN',
+            };
+        }
+
```
prastoin, line 350: Thoughts: user will then have to sign in using his new password ? What will do or display the client after this operation success ?

packages/server/app/Controllers/Http/AuthenticationController.test.ts
```diff
@@ -937,3 +939,159 @@ test.group('Request password reset', (group) => {
...
+        const parsedResponseBody = ValidatePasswordResetTokenResponseBody.parse(
+            response.body,
+        );
+
+        assert.deepStrictEqual(parsedResponseBody, {
+            status: 'INVALID_TOKEN',
+        });
+    });
+
+    test('Returns error if token has expired', async (assert) => {
```
prastoin, line 1056: praise: great coverage !

prastoin (approved):
Nice work !
Leaved few comments
Still ready to be merged

packages/server/app/Controllers/Http/AuthenticationController.ts
Devessier: This route will be requested at this step:

CleanShot 2022-04-04 at 09 48 55@2x

It's only meant to confirm that the token is valid. After that, we redirect the user to the form where she fills her new password.
And when the user submits her new password, we send to the backend the email, the token and the new password. And we'll authenticate her automatically if the operation succeeds!

PR #553, opened 2022-04-04
Create password reset token form screen

Devessier:


packages/client/screens/PasswordResetConfirmationTokenScreen.tsx
```diff
@@ -0,0 +1,221 @@
...
+                type: 'server',
+                message: 'Code is invalid.',
+            });
+        }, [passwordResetCodeIsInvalid, setError]);
+
+        function handlePasswordResetTokenSubmit({
+            code,
+        }: PasswordResetConfirmationTokenFormFieldValues) {
+            appService.send({
+                type: 'SUBMIT_PASSWORD_RESET_CONFIRMATION_FORM',
```
prastoin, line 56: thoughts: I'm not sure about this event name, unless I'm mistaken this event will perform a reset password token verification and not a password reset confirmation ( that, in my mind, could be the next step after validating a password reset code where user will have to define a new password )

packages/client/machines/appMachine.ts
```diff
@@ -229,66 +232,175 @@ export function createAppMachine({
...
+
+                                            actions: appModel.assign({
+                                                email: (_, event) =>
+                                                    event.email,
+                                            }),
+                                        },
+                                    },
+                                },
 
                                 requestingPasswordReset: {
```
prastoin, line 248: Praise: great work ! nice to see a such well managed authentication machine

packages/client/__tests__/AuthenticationPasswordReset.test.tsx
```diff
@@ -179,6 +182,154 @@ const passwordResetMachine =
...
+                                expect(
+                                    passwordResetCodeScreenContainer,
+                                ).toBeTruthy();
+                            });
+                        },
+                    },
+                    initial:
+                        'Displaying password reset successful request toast',
+                    states: {
+                        'Displaying password reset successful request toast': {
```
prastoin, line 202: Question: why are we expecting a toast when entering the password reset workflow ?

packages/client/machines/appMachine.ts
Devessier: Aha! Long live state machines!

packages/client/__tests__/AuthenticationPasswordReset.test.tsx
Devessier: I added a comment to explain the reason: https://github.com/AdonisEnProvence/MusicRoom/pull/553/commits/21dee2f90a6ebf7c1d3b40e4610ea61ce39d1cc7#diff-4982ec3e27a3d7f1d2d966a855028736a70f7956b076bf673165d46e9ebc89bcR202.

Please note the state is named _Password reset **code** form_. In this step we want to get the confirmation _code_.

prastoin (approved):
Great work ! ready to be merged
Thanks for the the tiny modifications

PR #554, opened 2022-04-05
Create route to change user password

Devessier:
Must be reviewed after https://github.com/AdonisEnProvence/MusicRoom/pull/553

### DIFF
--- a/packages/server/app/Controllers/Http/AuthenticationController.test.ts
+++ b/packages/server/app/Controllers/Http/AuthenticationController.test.ts
@@ -6,6 +6,10 @@ import {
     RequestPasswordResetRequestBody,
     RequestPasswordResetResponseBody,
     ResendConfirmationEmailResponseBody,
+    ResetPasswordRequestBody,
+    ResetPasswordResponseBody,
+    ResetPasswordSuccessfulApiAuthenticationResponseBody,
+    ResetPasswordSuccessfulWebAuthenticationResponseBody,
     SignInFailureResponseBody,
     SignInRequestBody,
     SignInSuccessfulApiTokensResponseBody,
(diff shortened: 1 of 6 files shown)


packages/server/app/Controllers/Http/AuthenticationController.ts
```diff
@@ -355,6 +357,55 @@ export default class AuthenticationController {
...
+        const isInvalidToken = isValidToken === false;
+        if (isInvalidToken === true) {
+            response.status(400);
+
+            return {
+                status: 'INVALID_TOKEN',
+            };
+        }
+
+        if ((await user.isSamePassword(password)) === true) {
```
prastoin, line 391: praise: great !

packages/server/app/Controllers/Http/AuthenticationController.ts
```diff
@@ -355,6 +357,55 @@ export default class AuthenticationController {
...
+        response,
+    }: HttpContextContract): Promise<ResetPasswordResponseBody> {
+        const { email, password, token, authenticationMode } =
+            ResetPasswordRequestBody.parse(request.body());
+
+        const user = await User.findBy('email', email);
+        if (user === null) {
+            response.status(400);
+
+            return {
```
prastoin, line 372: question: why would this error be an invalid token ?

packages/server/app/Controllers/Http/AuthenticationController.test.ts
```diff
@@ -1095,3 +1099,354 @@ test.group('Validate password reset token', (group) => {
...
+            .expect(200);
+        const parsedResponseBody =
+            ResetPasswordSuccessfulWebAuthenticationResponseBody.parse(
+                response.body,
+            );
+
+        assert.deepStrictEqual(parsedResponseBody, {
+            status: 'SUCCESS',
+        });
+
```
prastoin, line 1295: Suggestion: Unless I'm mistaken, you do not verify in your tests that the user password has changed ? We could compare at least that the hashed stored password has changed after the reset ?

packages/server/app/Models/User.ts
```diff
@@ -160,9 +170,11 @@ export default class User extends BaseModel {
         {
             token,
             tokenType,
+            revoke = false,
         }: {
             token: string;
             tokenType: TokenTypeName;
+            revoke?: boolean;
```
prastoin, line 177: Question: I'm not sure to understand what this method job's doing ? It's not only about reading value but also mutating it ? Maybe should we create a sperated method `RevokeLastToken` that takes a tokenType ?

I mean that it's not because you achieve to retrieve a valid token that it will be "used" which mean needs to be revoked on the fly. Something could fail during the process we might want to wait the operation end before revoking a related token ?

packages/server/app/Controllers/Http/AuthenticationController.ts
Devessier: Before this route is called, the email and the token pair must have been verified. I went to the simplest solution and decided that if email/token pair does not exist or is invalid, we return an `INVALID_TOKEN` error.

packages/server/app/Controllers/Http/AuthenticationController.test.ts
Devessier: You're right! I'm going to update the successful path test.

packages/server/app/Controllers/Http/AuthenticationController.ts
prastoin: It seems legit to me !

packages/server/app/Models/User.ts
Devessier: I can not agree more with you! I really went to the simplest solution.
The reason is that `checkToken` method does not return a `Token` but a boolean telling whether a token exists. We should replace this function with another one called like `getToken`, that actually returns a `Token`, to be able to revoke it **simply** afterwards.

If we keep `checkToken` function, we should create another function such as `revokeToken`, which would take a plain token as parameter, and that would have to iterate over all active tokens and compare their hashed value with the plain token.

Devessier: I think that a function that revokes _the last token_ could revoke another token than the one we want. We should better decide exactly which token we want to revoke.

prastoin: I see !
> I think that a function that revokes the last token could revoke another token than the one we want. We should better decide exactly which token we want to revoke.

Indeed there could be some race conditions :laughing:

packages/server/app/Controllers/Http/AuthenticationController.test.ts
```diff
@@ -1293,6 +1295,17 @@ test.group('Reset password', (group) => {
...
         });
 
+        await user.refresh();
+
+        invariant(
+            user.password !== undefined,
+            'The hashed password of the user must have been computed',
+        );
+        const hasPasswordChanged =
+            (await Hash.verify(user.password, newPlainPassword)) === true;
```
prastoin, line 1305: praise: great !

packages/server/app/Controllers/Http/AuthenticationController.test.ts
```diff
@@ -1430,6 +1430,54 @@ test.group('Reset password', (group) => {
         assert.isTrue(token.isRevoked);
     });
 
+    test('Does not revoke token if provided password is same as old password', async (assert) => {
```
prastoin, line 1433: Some TDD ? :eyes:

packages/server/app/Models/User.ts
```diff
@@ -165,18 +165,16 @@ export default class User extends BaseModel {
...
             token,
             tokenType,
-            revoke = false,
         }: {
             token: string;
             tokenType: TokenTypeName;
-            revoke?: boolean;
         },
-    ): Promise<boolean> {
+    ): Promise<Token | undefined> {
```
prastoin, line 177: Praise: great !

prastoin (approved):
Excellent work !
Ready to be merged
Well done !

Issue #544, opened 2022-03-28
Signup/Signin using a google account


prastoin:
After a google oauth authentication we will retrieve a user googleAccessToken that will be sent to the backend to be converted to private user information especially:
`googleID`
`google_email`
`google_user_name`
The backend will seach in database for a matching already existing user goolgeID.
It the operations succeed then we will perform a simple sign in if not we will perform a sign up.

We should keep in mind that only an authenticated user can bind/link a google account to her musicroom account.

client:
- [x] create a google auth button that will handle expo `auth-session` google auth logic
- [x] create a service to send a google authentication api or web depending on the platform
- [x] integ tests coverage
~~- [ ] e2e tests coverage~~

server:
- [x] create a `google-social-auth` route that takes an authenticationMode and an googleUserAccessToken
- [x] handle the `google-social-auth` perform a request using the received googleUserAcessToken to the `https://www.googleapis.com/oauth2/v3/userinfo` google api, from there exchange the googleAccessToken for user google email nickname and googleID.
If the a matching googleID is found in database sign in the user
If no mathcing googleID is found in database sign perform a classic sign up, using user google email ( beware of edge cases such as already existing user with received email )

PR #555, opened 2022-04-06
Client expo auth session refactor and tests

prastoin:
close https://github.com/AdonisEnProvence/MusicRoom/issues/544


packages/client/__tests__/GoogleAuthentication.test.tsx
```diff
@@ -0,0 +1,560 @@
...
+                        {
+                            target: 'Rendering signing in screen.Server user access token verification error.Unavailable related nickname or email',
+                        },
+                    'Make server send back an nickname nor email invalid error':
+                        {
+                            target: 'Rendering signing in screen.Server user access token verification error.Invalid related nickname or email',
+                        },
+                    'Make server send back an unknown error': {
+                        target: 'Rendering signing in screen.Server user access token verification error.Unknown server error',
+                    },
```
Devessier, line 220: **💭 thought:** ‏When dealing with events in `@xstate/test` that modify the setup, I prefer to dissociate them from actions that a user could actually do.

It means that I would not call `pressContinueWithGoogleButton()` in the same events that modify how the server responds. To keep tracks in the machine that the setup has changed, I would assign to the context a boolean indicating that, like here: https://github.com/AdonisEnProvence/MusicRoom/blob/f38450da3a4fa7924bbd3696ffec3709683963a7/packages/client/__tests__/AuthenticationPasswordReset.test.tsx#L39.

But, to be honest, there is a lot of possible cases. And writing all these booleans, and a guard for each, would be tedious!

Devessier (approved):
Really nice job!
I like the integration of Google OAuth with XState!
We can merge.

packages/client/tests/server/handlers.ts
```diff
@@ -935,10 +945,71 @@ export const handlers = [
...
+                ctx.json({
+                    type: 'success',
+                    errorCode: null,
+                    authentication,
+                    params: {},
+                    url: '',
+                    error: undefined,
                 }),
             );
         },
```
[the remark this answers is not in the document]
prastoin: Not me :eyes:

packages/client/__tests__/GoogleAuthentication.test.tsx
prastoin: Indeed you're completely right !
As you say a lot to deal with I went to the fast solution

PR #562, opened 2022-04-07
Allow to go back to previous screen on password reset wizard

Devessier:


packages/client/__tests__/AuthenticationPasswordReset.test.tsx
```diff
@@ -30,11 +31,16 @@ interface TestingContext {
     screen: ReturnType<typeof render>;
 }
 
-const VALID_PASSWORD_RESET_CODE = '123456';
+/**
```
prastoin, line 34: praise: tricky !

packages/client/machines/appMachine.ts
```diff
@@ -524,6 +524,36 @@ export function createAppMachine({
                                     },
                                 },
                             },
+
+                            /**
```
prastoin, line 528: Question: Is the user able to go back on every reset password wizard form screen ? event the set new password one ?

packages/client/__tests__/AuthenticationPasswordReset.test.tsx
```diff
@@ -1332,5 +1392,53 @@ cases<{
...
+                    },
+                    {
+                        type: 'Go back to previous screen',
+                    },
+                    {
+                        type: 'Request password reset',
+                    },
+                    {
+                        type: 'Type on password reset code field',
+                        code: VALID_PASSWORD_RESET_CODE_LIST[1],
```
prastoin, line 1429: The previous VALID_PASSWORD_RESET_CODE_LIST[0] is well destroyed in any state machines context ?

prastoin (approved):
Great work !
I added few questions  just to be sure to understand the business logic !
We can merge

packages/client/machines/appMachine.ts
Devessier: Yes, the user can go back from any screen about password reset.

packages/client/__tests__/AuthenticationPasswordReset.test.tsx
Devessier: The `passwordResetCode` property will be reassigned, yes. Otherwise the test would not work as `@xstate/test` would have to go back to a state it has already visited! Which is not possible.

PR #563, opened 2022-04-08
Ensure new password must be strong enough

Devessier:


packages/server/app/Controllers/Http/AuthenticationController.test.ts
```diff
@@ -1207,6 +1207,43 @@ test.group('Reset password', (group) => {
         });
     });
 
+    test('Returns an error if password is not strong enough', async () => {
```
prastoin, line 1210: Thought: On first ready I couldn't determine the password strength validity check context ( e.g: sign up, reset password )

prastoin (approved):
More coverage implies and deeper and better sleep

Issue #564, opened 2022-04-08
Reset password and sign in after google sign up


prastoin:
Things to test
- [ ] 1/ Reset password
A user that has been signing up using a google authentication should perform a password reset, it should then define a password as it was null before. The same user will perform a sign up using his googleInformation email and the freshly defined password, he should be able to sign in

- [ ] 2/ Sign in using empty password
A user that has been signing up using a google authentication should perform a sign in passing a null nor a empty string as a password.
The sign in should fail

prastoin:
Closing this issues as it has been resolved

PR #566, opened 2022-04-08
Ensure accounts created from Google can change their password

Devessier:


prastoin:
related to https://github.com/AdonisEnProvence/MusicRoom/issues/564

prastoin (approved):
Great way to improve the coverage without duplicating the tests
Jest in case has better watch out

PR #582, opened 2022-04-21
mpe test-e fixing race conditions

prastoin:
Test was failing mainly because we wanna add a track to the created mpe that has been used to created the current mtv room. Then the selector will stop on the mtv room mini music player instead of the mpe tracks search results
Also improved the test-e selectors 

also fixed e2e sign up tests issue where it sould target previous sign in screen input instead of the sign up one ! thanks @Devessier

### DIFF
--- a/packages/client/e2e/_utils/mock-http.ts
+++ b/packages/client/e2e/_utils/mock-http.ts
@@ -8,7 +8,10 @@ export interface KnownSearchesElement {
     duration: number;
 }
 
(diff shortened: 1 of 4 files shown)


Devessier (dismissed):
We can merge! It makes the e2e test mpe/test-e pass.
Nice job!

Devessier (approved):
Great job!
We can merge.

PR #604, opened 2022-04-28
fix(server): add DEPLOY_PACKAGE and build.sh script to run migration …

prastoin:
…for server deploy only


Devessier (approved):
Really interesting! We can merge.

Issue #613, opened 2022-04-29
Bugs in production


prastoin:
- [ ] Youtube player issue on IOS, does not start the iframe at all
We should try few solutions such as killing and respawing the iframe on each new track
- [x] on go to next track the iframe does not start the following see above
- [x] the delegation is laggy when a track is nexted
- [x] the user name is not trimmed at all
- [x] mtv room with constraint does not send creator position on creation

prastoin:
We couldn't fix the ios web browser reading video

PR #614, opened 2022-04-29
refactor(server): trim userNickname on nickname update

prastoin:
related to https://github.com/AdonisEnProvence/MusicRoom/issues/613


Devessier (approved):
Perfect job!
We can merge.

PR #615, opened 2022-04-29
Ask for location after set physical constraint in mtv creation form

prastoin:
related to https://github.com/AdonisEnProvence/MusicRoom/issues/613
close https://github.com/AdonisEnProvence/MusicRoom/issues/590


Devessier (approved):
A really cool demonstration that state machines are so nice!
We can merge.

PR #622, opened 2022-05-02
Apply separator between tracks only when necessary

Devessier:


prastoin (approved):
Nice one !
Ready to be merged

PR #627, opened 2022-05-09
Fix mpe creator as invited user

prastoin:
A creator could'nt edits his own mpe as inside the api thas is not explicitly tested the default value is isInvited = false


Devessier (approved):
We can merge.

